PR #51, opened 2015-05-18
Physics Helpers

Hoikas:
These are some GenericPhyical helpers that generate box and sphere bounds for us. Nice for korman.

### DIFF
--- a/core/Math/hsGeometry3.h
+++ b/core/Math/hsGeometry3.h
@@ -67,6 +67,17 @@ struct PLASMA_DLL hsVector3 {
     { return hsVector3(X + other.X, Y + other.Y, Z + other.Z); }
 
     /**
+     * Adds the vectors in place using vector addition:
+     * \f$[x_1,y_1,z_1] + [x_2,y_2,z_2] = [x_1+x_2,y_1+y_2,z_1+z_2]\f$
+     */
+    hsVector3& operator+=(const hsVector3& other) {
+        X += other.X;
+        Y += other.Y;
+        Z += other.Z;
+        return *this;
+    }
+
+    /**
      * Subtracts the vectors using vector addition:
      * \f$[x_1,y_1,z_1] - [x_2,y_2,z_2] = [x_1-x_2,y_1-y_2,z_1-z_2]\f$
      */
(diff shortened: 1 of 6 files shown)


core/Math/hsGeometry3.h
```diff
@@ -67,6 +67,16 @@ struct PLASMA_DLL hsVector3 {
     { return hsVector3(X + other.X, Y + other.Y, Z + other.Z); }
 
     /**
+     * Adds the vectors in place using vector addition:
+     * \f$[x_1,y_1,z_1] + [x_2,y_2,z_2] = [x_1+x_2,y_1+y_2,z_1+z_2]\f$
+     */
+    void operator+=(const hsVector3& other) {
```
zrax: `operator+=` should generally return a reference to `*this`

core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -733,3 +735,39 @@ void plGenericPhysical::setTMDBuffer(size_t tmdSize, const unsigned char* tmdBuf
...
+    // PhysX and ODE can do true boxes
+    fDimensions = (bounds.getMaxs() - bounds.getMins()) * .5f;
+    fOffset = bounds.updateCenter();
+
+    // Havok wants actual geometry
+    fVerts.resize(8);
+    fVerts[0] = fVerts[1] = fVerts[2] = fVerts[4] = bounds.getCorner();
+    fVerts[1] += bounds.getAxis(0);
+    fVerts[2] += bounds.getAxis(1);
+    fVerts[4] += bounds.getAxis(2);
```
zrax: This seems ... odd...  Are the axes really keeping track of the size of the bounds?

Hoikas: It's been awhile since I wrote this code, so that's a good question. At one point, I hacked in code to generate faces for these verts and viewed the trimesh proxy in CWE, and the resulting geometry matched the visuals. Of course, that could have just been a fluke. IIRC I ported this from some code in CWE that renders bounding boxes.

core/Math/hsGeometry3.h
Hoikas: Le d'oh

core/PRP/Physics/plGenericPhysical.cpp
Hoikas: This looks like the correct way to do it for unaligned BBs, based off CWE's hsBounds3Ext's GetCorners(). I'm going to add that to libhsplasma for sanity's sake and remove these hacks.

Hoikas: Also, AFAICT, in libhsplasma, the dists are not actually updated.

Hoikas:
Fixed issues from review

PR #52, opened 2015-06-04
Moar PyHSPlasma Bindings

Hoikas:
Lots of logic-type stuff that will enable Korman to do some very basic things like panic link and footstep regions.

### DIFF
--- a/Python/PRP/ConditionalObject/pyConditionalObject.cpp
+++ b/Python/PRP/ConditionalObject/pyConditionalObject.cpp
@@ -1,0 +1,153 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * HSPlasma is distributed in the hope that it will be useful,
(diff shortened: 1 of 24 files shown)


Python/PRP/ConditionalObject/pyConditionalObject.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+extern "C" {
+
+static PyObject* pyConditionalObject_new(PyTypeObject* type, PyObject* args, PyObject* kwds) {
+    PyErr_SetString(PyExc_RuntimeError, "plConditionalObject is abstract");
+    return NULL;
+}
+
+static PyObject* pyConditionalObject_getSatisfied(pyConditionalObject* self, void*) {
+    return PyBool_FromLong(IConvertCond(self)->getSatisfied() ? 1 : 0);
```
zrax: Why are the IConvertCond calls necessary?  self->fThis should already return a plConditionalObject...

Hoikas: With VC++, if you try to access fThis by casting self to anything but its final type, there is an access violation. IIRC Branan came up with this solution many moons ago. It's all over PyHSPlasma, I just chose to shorten things a bit here because the long version is just silly.

Hoikas:
Updated to fix an issue with accessing new object types from pyKey.object

PR #61, opened 2015-07-04
PyHSPlasma Animation Improvements

Hoikas:
Improves the animation bits of PyHSPlasma by adding correct duck-typing for many return values and fixes many of those pesky plCreatable casts. This doesn't fix every single problem in the land, but it hits all of the ones needed for exporting simple animations in Korman and then some :smile:.

### DIFF
--- a/Python/PRP/Animation/pyLeafController.cpp
+++ b/Python/PRP/Animation/pyLeafController.cpp
@@ -129,7 +129,6 @@ static int pyLeafController_setEaseControllers(pyLeafController* self, PyObject*
             PyErr_SetString(PyExc_TypeError, "easeControllers should be a list of plEaseControllers");
             return -1;
         }
-        ((pyEaseController*)itm)->fPyOwned = false;
         controllers[i] = ((pyEaseController*)itm)->fThis;
     }
     IConvertController(self)->setEaseControllers(controllers);
--- a/Python/PRP/Animation/pyLeafController.cpp
+++ b/Python/PRP/Animation/pyLeafController.cpp
@@ -68,7 +68,6 @@ static PyObject* pyLeafController_setKeys(pyLeafController* self, PyObject* args
             PyErr_SetString(PyExc_TypeError, "setKeys expects a list of hsKeyFrames and an int");
             return NULL;
         }
-        ((pyKeyFrame*)itm)->fPyOwned = false;
         keys[i] = ((pyKeyFrame*)itm)->fThis;
     }
     IConvertController(self)->setKeys(keys, type);
--- a/Python/PRP/Animation/pyLeafController.cpp
+++ b/Python/PRP/Animation/pyLeafController.cpp
@@ -50,31 +50,6 @@ static PyObject* pyLeafController_hasEaseControllers(pyLeafController* self) {
     return PyBool_FromLong(IConvertController(self)->hasEaseControllers() ? 1 : 0);
 }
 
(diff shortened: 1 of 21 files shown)


Python/PRP/Animation/pyLeafController.cpp
```diff
@@ -64,28 +68,27 @@ static PyObject* pyLeafController_setKeys(pyLeafController* self, PyObject* args
             PyErr_SetString(PyExc_TypeError, "setKeys expects a list of hsKeyFrames and an int");
             return NULL;
         }
-        ((pyKeyFrame*)itm)->fPyOwned = false;
```
zrax: Should this have been removed?  I would expect the controller to take ownership of the keys...

core/PRP/Animation/plLeafController.cpp
```diff
@@ -441,7 +441,7 @@ void plLeafController::setKeys(const std::vector<hsKeyFrame*>& keys, unsigned in
     DeallocKeys();
     AllocKeys(keys.size(), type);
     for (size_t i=0; i<keys.size(); i++)
-        fKeys[i] = keys[i];
+        *fKeys[i] = *keys[i];
```
zrax: Good catch...  Looks like it also needs to be fixed for `setEaseControllers` though

Python/PRP/Animation/pyLeafController.cpp
Hoikas: :wink:

core/PRP/Animation/plLeafController.cpp
Hoikas: I think this is mirroring Cyan's implementation somewhat literally. IMO copying the keys is the correct solution in this case--the keys could be coming from another controller, adding the gotcha of one controller stealing another controller's owned keys. There's also the fPyOwned gotcha where we have a TypeError raised and some keys are leaked as a result.

zrax: Ah, good points.  But we still need `setEaseControllers` to get fixed as well :)

Hoikas: indeed

zrax:
While you're at it (sorry :P), the keys and controllers for `pyLeafController` are somewhat inconsistent...  Calling `leaf.keys = foo` generates an error and recommends `leaf.setKeys(foo)`, but `leaf.easeControllers` is directly assignable...  It would be nice if these were consistent >.>

Hoikas:
I was thinking the same thing. Do we want to get rid of setKeys entirely?

zrax:
Yeah, I'd be fine with just using the `.keys =` and `.easeControllers =` methods, and removing `setKeys` from Python.

Hoikas:
Updated to make pyLeafController.keys more awesome. I also implemented pyAGMasterMod and pyAGModifier while you were looking away :tongue:

Python/PRP/Animation/pyLeafController.cpp
```diff
@@ -101,13 +82,44 @@ static PyObject* pyLeafController_getEaseControllers(pyLeafController* self, voi
 }
 
 static int pyLeafController_setType(pyLeafController* self, PyObject* value, void*) {
-    PyErr_SetString(PyExc_RuntimeError, "To set the key type, use setKeys");
+    PyErr_SetString(PyExc_RuntimeError, "To set the key type, use the keys setter");
```
zrax: Ah, that explains the inconsistency...

PR #65, opened 2015-07-12
pyTimerCallbackMsg & pyEnableMsg

Hoikas:
Python bindings for these useful messages in Responders.

### DIFF
--- a/Python/PRP/Message/pyTimerCallbackMsg.cpp
+++ b/Python/PRP/Message/pyTimerCallbackMsg.cpp
@@ -1,0 +1,150 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * HSPlasma is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


Python/PRP/Message/pyTimerCallbackMsg.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (value == NULL || !PyFloat_Check(value)) {
+        PyErr_SetString(PyExc_TypeError, "time expects a float");
+        return -1;
+    }
+    self->fThis->setTime((float)PyFloat_AsDouble(value));
+    return 0;
+}
+
+static PyGetSetDef pyTimerCallbackMsg_GetSet[] = {
+    { _pycs("ID"), (getter)pyTimerCallbackMsg_getID, (setter)pyTimerCallbackMsg_setID, NULL, NULL },
```
zrax: Why upper case?

Hoikas: I was trying to be consistent with other parts of the code I knew about, such as pyVector3's upper case X, Y, and Z when the C++ variable was upper as well. I couldn't find any instances of ID or id in PyHSPlasma to offer guidance here, unfortunately.

zrax: Fair enough

PR #69, opened 2015-07-19
Implement plFollowMod and pyFollowMod.

Deledrius:


Python/PRP/pyCreatableConvert.cpp
```diff
@@ -377,7 +379,7 @@ plCreatable* IConvert(pyCreatable* pCre)
     else if (Py_TYPE(pCre) == &pyResponderModifier_Type) return dynamic_cast<plCreatable*>(reinterpret_cast<plResponderModifier*>(pCre->fThis));
     //else if (Py_TYPE(pCre) == &pyAxisAnimModifier_Type) return dynamic_cast<plCreatable*>(reinterpret_cast<plAxisAnimModifier*>(pCre->fThis));
     //else if (Py_TYPE(pCre) == &pyLayerLightBase_Type) return dynamic_cast<plCreatable*>(reinterpret_cast<plLayerLightBase*>(pCre->fThis));
-    //else if (Py_TYPE(pCre) == &pyFollowMod_Type) return dynamic_cast<plCreatable*>(reinterpret_cast<plFollowMod*>(pCre->fThis));
+    else if (Py_TYPE(pCre) == &pyFollowMod_Type) return dynamic_cast<plCreatable*>(reinterpret_cast<plFollowMod*>(pCre->fThis));
```
Hoikas: Needs to be added to ICreate as well, otherwise trying to access the object from a plKey will return a plCreatable

core/PRP/Modifier/plFollowMod.h
```diff
@@ -0,0 +1,66 @@
...
+        kRotate = 0x8,
+        kFullTransform = kPosition | kRotate
+    };
+
+    enum FollowLeaderType {
+        kFollowPlayer, kFollowObject, kFollowCamera, kFollowListener
+    };
+
+protected:
+    FollowLeaderType    fLeaderType;
```
Hoikas: that's kind of evil

Deledrius: Check out [PRP/Animation/plLineFollowMod.h] (https://github.com/H-uru/libhsplasma/blob/master/core/PRP/Animation/plLineFollowMod.h#L44). :expressionless:

Hoikas: ++evil

Python/PRP/pyCreatableConvert.cpp
Deledrius: I had a feeling I was forgetting one of the several places to which we add a new object.

core/PRP/Modifier/plFollowMod.h
Deledrius: Oh, and [Plasma/FeatureLib/pfAnimation/plFollowMod.h] (https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/FeatureLib/pfAnimation/plFollowMod.h#L78). :see_no_evil:

Python/PRP/pyCreatableConvert.cpp
Deledrius: Done.

core/PRP/Modifier/plFollowMod.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+void plFollowMod::IPrcWrite(pfPrcHelper* prc) {
+    plSingleModifier::IPrcWrite(prc);
+
+    prc->startTag("Flags");
+    prc->writeParamHex("leaderType", static_cast<uint8_t>(fLeaderType));
+    prc->writeParamHex("mode", fMode);
+    prc->endTag(true);
+
+    if (fLeaderType == kFollowObject) {
```
zrax: If the key is written unconditionally in the `read` and `write` methods, I'd prefer that it was unconditionally in the PRC as well, even if it Should Always Be Empty <sup>TM</sup>.

Deledrius: Good point.

core/PRP/Modifier/plFollowMod.cpp
```diff
@@ -0,0 +1,60 @@
...
+    S->writeByte(fLeaderType);
+    S->writeByte(fMode);
+    mgr->writeKey(S, fLeader);
+}
+
+void plFollowMod::IPrcWrite(pfPrcHelper* prc) {
+    plSingleModifier::IPrcWrite(prc);
+
+    prc->startTag("Flags");
+    prc->writeParamHex("leaderType", static_cast<uint8_t>(fLeaderType));
```
zrax: `leaderType` is a sequential enumeration, so it doesn't need to be written as hex.  Ideally, the names should be marshalled (some other PRC types do this), but I'll accept it without if you don't want to add that :)

Deledrius: I considered it as I'd prefer named flags in the PRC, but it's a lot more work and not strictly necessary for the interface.

PR #70, opened 2015-07-24
Soft Volumes

Hoikas:
Disclaimer: the pyVolumeIsect support is lazy.

### DIFF
--- a/core/Math/hsGeometry3.h
+++ b/core/Math/hsGeometry3.h
@@ -109,14 +109,7 @@ struct PLASMA_DLL hsVector3 {
     }
 
     /** Converts this vector a unit vector such that its magnitude is 1 */
-    void normalize() {
-        float length = magnitude();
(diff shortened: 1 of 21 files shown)


core/Math/hsGeometry3.h
```diff
@@ -107,6 +107,16 @@ struct PLASMA_DLL hsVector3 {
                          (Z * other.X) - (X * other.Z),
                          (X * other.Y) - (Y * other.X));
     }
+
+    /** Converts this vector a unit vector such that its magnitude is 1 */
+    void normalize() {
+        float length = magnitude();
+        if (length == 0.f || length == 1.f)
```
zrax: Floating point math is not precise...  I feel like we should have a "close enough to 1.f" rather than "exactly 1.f" check, or re-normalizing may become lossy

Hoikas: Ok, I think I've done it correctly?

PR #71, opened 2015-11-30
Deal with python library version '2.7.10+'

diafero:
Without this patch, I am getting an error, the interpreter search complaining that there is no version '2.7.10+'. I have no idea who thought it was a good idea to make *that* the version number returned by the library search.


Deledrius:
:+1: 
This is also useful for `Python 3.5.1+`.

PR #72, opened 2016-01-25
Improved AgeLink Bindings

Hoikas:
Improved the pyAgeLinkStruct and pyAgeInfoStruct bindings and added pySpawnPointInfo bindings such that Korman can generate plLinkToAgeMsg in Python. There were many bugs in the previous pyAgeLinkStruct and pyAgeInfoStruct bindings that have been corrected, such as spurious error values and illegal return types.


Python/PRP/Misc/pyAgeInfoStruct.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+static int pyAgeInfoStruct_setAgeInstanceGuid(pyAgeInfoStruct* self, PyObject* value, void*) {
+    if (value == NULL || !PyAnyStr_Check(value)) {
+        PyErr_SetString(PyExc_TypeError, "ageInstanceGuid should be a string");
+        return -1;
+    }
+    try {
+        plUuid uuid(PyStr_To_PlStr(value));
+        self->fThis->setAgeInstanceGuid(uuid);
+    } catch (hsBadParamException&) {
```
zrax: should probably be `const` reference...

Hoikas: Done

PR #73, opened 2016-02-01
Python Sound Bindings

Hoikas:
Enough bindings to (theoretically) export sounds from Korman.

### DIFF
--- a/Python/PRP/Audio/pyFadeParams.cpp
+++ b/Python/PRP/Audio/pyFadeParams.cpp
@@ -135,7 +135,7 @@ static PyGetSetDef pyFadeParams_GetSet[] = {
 
 PyTypeObject pyFadeParams_Type = {
     PyVarObject_HEAD_INIT(NULL, 0)
-    "PyHSPlasma.plFadeParams",          /* tp_name */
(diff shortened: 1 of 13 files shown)


Python/PRP/Audio/pyFadeParams.cpp
```diff
@@ -0,0 +1,215 @@
...
+    { _pycs("fadeSoftVol"), (getter)pyFadeParams_getFadeSoftVol,
+      (setter)pyFadeParams_setFadeSoftVol, NULL, NULL },
+    { _pycs("currTime"), (getter)pyFadeParams_getCurrTime,
+      (setter)pyFadeParams_setCurrTime, NULL, NULL },
+    { NULL, NULL, NULL, NULL, NULL }
+};
+
+PyTypeObject pyFadeParams_Type = {
+    PyVarObject_HEAD_INIT(NULL, 0)
+    "PyHSPlasma.plFadeParams",          /* tp_name */
```
zrax: Should be `PyHSPlasma.plSound.plFadeParams` based on where the type is instantiated.

Hoikas: Fixed!

PR #74, opened 2016-02-15
Implement pyDistOpacityMod and pyFadeOpacityMod.

Deledrius:
More bindings needed for Korman.


Hoikas:
:+1:

zrax:
It seems odd to have a different API for Python vs. C++, but that's something to worry about another day.

Deledrius:
I don't think it's that odd.  C++ and Python have different paradigms, so having an API that differs in places should not only be expected but encouraged as long as it makes sense to do so.

PR #75, opened 2016-03-05
Moar Sound Bindings

Hoikas:
This implements `pySoundMsg` and allows for `pyWAVHeader` objects to be constructed in Python--it's a really useful struct for exporters.


Python/PRP/Message/pySoundMsg.cpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+static PyObject* pySoundMsg_setCmd(pySoundMsg* self, PyObject* args) {
+    int cmd;
+    int value = 1;
+    if (!PyArg_ParseTuple(args, "i|i", &cmd, &value)) {
+        PyErr_SetString(PyExc_TypeError, "setCmd expects an int and an optional int");
+        return NULL;
+    }
+    self->fThis->setCmd(cmd, value != 0);
```
zrax: It would be handy if `plSoundMsg::setCmd` also defaulted to `value=true`, to match the Python behavior

Python/PRP/Audio/pySound.cpp
```diff
@@ -297,9 +297,9 @@ static PyGetSetDef pySound_GetSet[] = {
       (setter)pySound_setMinFalloff, NULL, NULL },
     { _pycs("outerVol"), (getter)pySound_getOuterVol,
       (setter)pySound_setOuterVol, NULL, NULL },
-    { _pycs("outerVol"), (getter)pySound_getInnerCone,
+    { _pycs("innerCone"), (getter)pySound_getInnerCone,
       (setter)pySound_setInnerCone, NULL, NULL },
-    { _pycs("outerVol"), (getter)pySound_getOuterCone,
+    { _pycs("outerCone"), (getter)pySound_getOuterCone,
```
zrax: :trollface:

Hoikas:
Added default value of true for plSoundMsg::setCmd

PR #76, opened 2016-04-12
Port to string_theory

zrax:
One string library to rule them all.

This was actually the first one I did, so there are some new APIs and shortcuts I probably missed, but I was tired of looking at strings >.>


zrax:
@Hoikas I'm especially interested to see how this affects korman, both functionality-wise and performance-wise (if at all).

Hoikas:
I'll try this one after #81 is merged

PR #77, opened 2016-06-21
Add a stub for pyRAMStream.close()

Hoikas:
I know this doesn't match the C++ API, but it seems more pythonic to be able to write:
```python
stream.close()
```
as a catch-all instead of
```python
if isinstance(stream, hsFileStream):
    stream.close()
```
when you're juggling both hsFileStream and hsRAMStream.

### DIFF
--- a/Python/Stream/pyEncryptedStream.cpp
+++ b/Python/Stream/pyEncryptedStream.cpp
@@ -56,12 +56,6 @@ static PyObject* pyEncryptedStream_open(pyEncryptedStream* self, PyObject* args)
     }
 }
 
-static PyObject* pyEncryptedStream_close(pyEncryptedStream* self) {
-    self->fThis->close();
-    Py_INCREF(Py_None);
-    return Py_None;
-}
-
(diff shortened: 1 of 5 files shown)


zrax:
Seems like close should be a virtual function in `hsStream` >.>

Hoikas:
I could refactorz it to be that way if you prefer.

zrax:
Yeah, I think that would be a better API.  Then we don't end up having to fix it for C++ too some day...

Hoikas:
And done!

core/Stream/hsStdioStream.h
```diff
@@ -28,7 +28,6 @@ class PLASMA_DLL hsStdioStream : public hsFileStream {
     explicit hsStdioStream(bool useStdErr = false);
 
     virtual bool open(const char*, FileMode) { return false; }
-    virtual void close() { }
```
zrax: This removal is incorrect.  The base class implements close() to close the associated file, which this overrides to remove.

Hoikas:
Removed an improper removal. :trollface:

PR #79, opened 2016-08-07
Static build

NadnerbD:
Added an option to build static libraries. (or rather, added an option to build shared libs, and set it to default on)

Also, I built this with MinGW 4.9.2, and it requires a newer std flag, or else it complains about some of the Win32 includes. Using gnu++11 shouldn't cause issues with other gcc environments, but it may be worth checking, especially given that mismatching this flag _will_ cause linking issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ endif()
 option(ENABLE_PYTHON "Build HSPlasma Python integration" OFF)
 option(ENABLE_TOOLS "Build the HSPlasma tools" ON)
 option(ENABLE_NET "Build HSPlasmaNet" ON)
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 
 try_compile(HAVE_OVERRIDE ${PROJECT_BINARY_DIR}
             ${PROJECT_SOURCE_DIR}/cmake/check_override.cpp)
--- a/Python/CMakeLists.txt
+++ b/Python/CMakeLists.txt
@@ -538,7 +538,7 @@ set(PYTHON_HDRS
     PyPlasma.h
 )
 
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG ${CMAKE_CXX_FLAGS_DEBUG}")
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR
         CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
     set(WARNING_FLAGS "-Wall -Wextra -Wno-unused-parameter")
-    set(CMAKE_CXX_FLAGS "-std=c++0x ${WARNING_FLAGS} ${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "-std=gnu++11 ${WARNING_FLAGS} ${CMAKE_CXX_FLAGS}")
```
zrax: I'm curious what problem this solved...  I'd prefer to stick to standard C++ with no GNU extensions if possible.  Also, c++0x should be a synonym for c++11, but the latter is only supported by GCC 4.7 or later (we're currently targetting GCC 4.6 minimum)

zrax: I pushed some changes to Trunk which should fix the MinGW issues without needing to change the `-std=` flag.  You may need additional adjustments for a static build though, since it fixes the incorrect detection on the `PLASMA_DLL` defines too -- but I haven't tested that.

NadnerbD:
Everything seems to work fine with your changes, so this is now just the addition of a static build.

PR #80, opened 2016-12-12
Pf camera python bindings

Lunanne:
I'm not sure this is ready to merge yet but I'm looking for some feedback/have some questions:
* Is the current work correct?
* Do I only need to add bindings for any camera subclass with extra values or does inheritance not work?
* How do I even test this?


.gitignore
```diff
@@ -26,3 +26,6 @@ Misc/TypeMap
 /dat
 /doc
 /*.kdev4
+
+/build/*
```
zrax: I usually don't add `/build` here any more, since it's a generic name that isn't created by any specific automatic build system (that I know of), and most will instead generate several build directories with different names.  Instead, you can generate your cmake build out-of-source, or create a .gitignore file in the build directory with only a `*` line as content.

Python/Math/pyVector3.cpp
```diff
@@ -381,4 +381,11 @@ PyObject* pyVector3_FromVector3(const hsVector3& vec) {
     return (PyObject*)pv;
 }
 
+hsVector3* pyVector3_AsVector3(PyObject* value)
+{
+	if(pyVector3_Check(value))
+	{
+		return ((pyVector3*)value)->fThis;
+	}
```
zrax: Need to return `NULL` if the check fails

Python/PRP/Camera/pyCameraBrain1.cpp
```diff
@@ -0,0 +1,369 @@
...
+#include <PyPlasma.h>
+#include <Math/pyGeometry3.h>
+#include <Util/pyBitVector.h>
+#include <PRP/Camera/plCameraBrain.h>
+#include "PRP/pyCreatable.h"
+#include "pyCameraBrain.h"
+
+extern "C" {
+
+	static PyObject* pyCameraBrain1_new(PyTypeObject* type, PyObject* args, PyObject* kwds) {
```
zrax: libhsplasma doesn't indent `extern "C"` blocks

Python/PRP/Camera/pyCameraBrain1.cpp
```diff
@@ -0,0 +1,369 @@
...
+		return 0;
+	}
+
+	static int pyCameraBrain1_setPOAOffset(pyCameraBrain1* self, PyObject* value, void*)
+	{
+		if (value == NULL || !pyVector3_Check(value)) {
+			PyErr_SetString(PyExc_TypeError, "camera POAOffset should be a vector3");
+			return -1;
+		}
+		self->fThis->setPOAOffset(*(pyVector3_AsVector3(value)));
```
zrax: This is essentially performing the `pyVector3_Check` twice (which is why I usually access it directly everywhere else:  `((pyVector3*)value)->fThis`).  If you want to add a wrapper (this is a good thing), you may want to remove the duplicate check -- that is, store the vector pointer off, check it for `NULL`, and print the error in the failed case (or set it in the successful case).

Python/Math/pyVector3.cpp
```diff
@@ -381,4 +381,11 @@ PyObject* pyVector3_FromVector3(const hsVector3& vec) {
     return (PyObject*)pv;
 }
 
+hsVector3* pyVector3_AsVector3(PyObject* value)
+{
+	if(pyVector3_Check(value))
```
zrax: Whitespace nit:  Space after `if` keyword

Python/Util/pyBitVector.cpp
```diff
@@ -273,4 +273,11 @@ PyObject* pyBitVector_FromBitVector(class hsBitVector& vec) {
     return (PyObject*)bv;
 }
 
+hsBitVector* pyBitVector_AsBitVector(PyObject* value)
```
zrax: Same comments for `pyVector3` apply here (minus whitespace nit)

core/PRP/Camera/plCameraBrain.h
```diff
@@ -79,6 +79,38 @@ class PLASMA_DLL plCameraBrain1 : public virtual plCameraBrain {
...
+	float getPOAAcceleration() const { return fPOAAccel; }
+	float getPOADeceleration() const { return fPOADecel; }
+	hsVector3 getPOAOffset() const { return fPOAOffset; }
+	float getXPanLimit() const { return fXPanLimit; }
+	float getZPanLimit() const { return fZPanLimit; }
+	float getPanSpeed() const { return fPanSpeed; }
+	float getZoomRate() const { return fZoomRate; }
+	float getZoomMax() const { return fZoomMax; }
+	float getZoomMin() const { return fZoomMin; }
+	hsBitVector getFlags() const { return fFlags; }
```
zrax: This should probably return a mutable (and immutable) reference rather than having get/set

Python/PRP/Camera/pyCameraBrain1.cpp
```diff
@@ -0,0 +1,369 @@
...
+		self->fThis->setEoAFlags(*(pyBitVector_AsBitVector(value)));
+		return 0;
+	}
+
+	static PyGetSetDef pyCameraBrain1_GetSet[] = {
+		{ _pycs("velocity"), (getter)pyCameraBrain1_getVelocity,
+		(setter)pyCameraBrain1_setVelocity, NULL, NULL },
+		{ _pycs("acceleration"), (getter)pyCameraBrain1_getAcceleration,
+		(setter)pyCameraBrain1_setAcceleration, NULL, NULL },
+	{ _pycs("deceleration"), (getter)pyCameraBrain1_getDeceleration,
```
Hoikas: troll indentation

Python/Util/pyBitVector.cpp
Hoikas: This should not be done at all. The other PyHSPlasma classes provide methods for setting and clearing bit flags instead of implementing a has-a member.

Hoikas:
Good job cranking out a ton of annoying boilerplate. A few things...

1. It looks like you did not add `plCameraBrain1` to the module definition in Module.cpp. This binding won't be available to Python if you don't do that.

2. Your code will crash with an access violation on Windows. `plCameraBrain1` is a base class for many other camera brains. There is an issue that @branan knows way more about than I do, when any derived pyKeyedObject tries to access `self->fThis` of the base class, bad things happen. You must therefore cast all access to `self->fThis`. This is the basic pattern I used in pyActivatorConditionalObject:

```
static inline plActivatorConditionalObject* IConvertCond(pyActivatorConditionalObject* self) {
    return plActivatorConditionalObject::Convert(IConvert((pyCreatable*)self));
}

static PyObject* pyActivatorConditionalObject_clearActivators(pyActivatorConditionalObject* self) {
    IConvertCond(self)->clearActivators();
    Py_INCREF(Py_None);
    return Py_None;
}
```

3. As mentioned above, there are lots of classes derived from `plCameraBrain1`, many of which have no additional added members. These would be quite easy to make bindings for very quickly.

4. Look at some of the other classes like `pyObjInterface` and other users of `pyVector3` and `pyBitVector` and try to match their implementation. One of the most annoying things I have encountered in developing Korman is when the libHSPlasma interface differs for no readily apparent reason. It makes the whole "duck typing" thing pointless when general conventions are (slightly) different :sad:

core/PRP/Camera/plCameraBrain.h
Lunanne: I'm not sure I fully understand, getFlags() should return a non-const reference? Should getPOAOffset work the same?

Python/CMakeLists.txt
```diff
@@ -171,6 +171,14 @@ set(PRP_AVTR_HDRS
     PRP/Avatar/pySittingModifier.h
 )
 
+set(PRP_CAMERA_SRCS
```
Deledrius: The [CMakeLists.txt for the `core` library](https://github.com/Lunanne/libhsplasma/blob/master/core/CMakeLists.txt#L155) uses **CMRA** for this.  I think we should probably use it here too for consistency.

zrax: I'm not actually a fan of those abbreviated names, but that's probably an issue for another PR

zrax:
This should probably be rebased on the recent python refactors -- it should simplify the boilerplate quite a bit.

Python/CMakeLists.txt
Deledrius: I agree. ;)

Hoikas:
Closed by #100

PR #81, opened 2016-12-13
Python bindings refactor (part 1)

zrax:
There's still a lot more I'd like to add to this, but this is a good start and should (slightly) reduce the amount of boilerplate and bad excessive re-casting needed for new PyHSPlasma additions.

Features
* Gets rid of virtual inheritance, which also implies eliminating the `plWhatever::Convert(IConvert((pyWhatever*)self))` crap.
* Adds `IConvert` and related conversions in several places where it was missing, to get a properly derived Plasma type
* Simplify and standardize a subset of the boilerplate a bit with macros.

Future directions (not in this PR, but I'm either working on or considering):
* Simplify properties and conversions with more macros
* Simplify declaring large data structures
* Align properties with C++ methods more closely (e.g. `setMin`, `setMax` instead of `setRange(min,max)`)
* Switch all currently value-type structures to references, so they can be used interchangeably (not as simple as it sounds; e.g. value types need a way to perform a deep-copy, which is currently missing).


zrax:
@Hoikas I'd like to ensure this doesn't break anything in Korman

Python/PRP/pyCreatableConvert.cpp
```diff
@@ -842,7 +471,8 @@ PyObject* ICreate(plCreatable* pCre)
         case kSoundMsg: pySoundMsg_FromSoundMsg(plSoundMsg::Convert(pCre));
         default:
             // many messages are not implemented, make sure they are at least a plMessage
-            if (dynamic_cast<plMessage*>(pCre)) return pyMessage_FromMessage(plMessage::Convert(pCre));
+            if (dynamic_cast<plMessage*>(pCre))
+                return pyMessage_FromMessage(plMessage::Convert(pCre));
             return pyCreatable_FromCreatable(pCre);
```
Hoikas: I can't help but wonder if there is some way to change all this crap into some kind of templatey goodness. It's pretty crazy that we have to have this kind of thing done by hand 😿

zrax:
The crashing is why I wanted you to test it :).  I figure you've got something closer to testable than me at the moment, since I've never tried to build korman.  Additionally, I don't know of any currently active tools that require the virtual inheritance.  And as my commit comment pointed out, I'd rather not encourage its use either due to the issues it has, and the fact that it wasn't even complete.

Python/PRP/pyCreatableConvert.cpp
zrax: Yeah, as I mentioned, there's still a lot more I want to do...  But the branch was sitting around gathering dust for a while, so I thought I'd get what I had out there for now and come back to add more later.

Hoikas:
I think the virtual inheritance was probably used in PlasmaClient. I'll see about testing these changes with Korman over my Christmas break.

PR #82, opened 2016-12-28
Generic PXPhysical

Hoikas:
This matches the proposed format for generic triangle mesh and convex hulls for CWE's plPXPhysical. See also H-uru/Plasma#533

### DIFF
--- a/core/PRP/Physics/plGenericPhysical.cpp
+++ b/core/PRP/Physics/plGenericPhysical.cpp
@@ -124,6 +124,9 @@ void plGenericPhysical::read(hsStream* S, plResManager* mgr) {
 void plGenericPhysical::write(hsStream* S, plResManager* mgr) {
     plSynchedObject::write(S, mgr);
 
+    if (fIndices.size() % 3 != 0)
+        throw hsBadParamException(__FILE__, __LINE__, "Indices must be triangles");
+
     PhysType wtype;
     if (S->getVer().isUniversal()) {
         wtype = fInternalType;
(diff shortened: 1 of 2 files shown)


core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -547,10 +547,31 @@ void plGenericPhysical::IReadPXPhysical(hsStream* S, plResManager* mgr) {
...
-        PXCookedData::readConvexMesh(S, this);
-    } else {    // Proxy or Explicit
-        PXCookedData::readTriangleMesh(S, this);
+    } else {
+        char tag[4];
+        S->read(4, tag);
+
+        if (memcmp(tag, "NXS\x01", 4) == 0) {
+            // Cooked PhysX stream
+            if (fBounds == plSimDefs::kHullBounds)
```
zrax: Do we need to rewind the stream by 4 bytes before reading the cooked meshes here?
EDIT: Nevermind, I see you deleted the header read below.  Leaving this comment in to remind myself.

core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -717,7 +741,13 @@ void plGenericPhysical::IWritePXPhysical(hsStream* S, plResManager* mgr) {
         if (!NxCookTriangleMesh(triDesc, buf))
             throw hsBadParamException(__FILE__, __LINE__, "Incorrect data for a PhysX Trimesh Bake");
 #else
-        throw hsNotImplementedException(__FILE__, __LINE__, "PhysX TriangleMesh");
+        S->write(4, "HSP\x01");
+        S->writeInt(fVerts.size());
+        for (size_t i = 0; i < fVerts.size(); ++i)
+            fVerts[i].write(S);
+        S->writeInt(fIndices.size() / 3);
```
zrax: Should we add a check somewhere that the indices array's size is valid (`%3 == 0`)?

Hoikas: That's probably a good idea. We'd want to do the same for the HKPhysical reader/writer because it serializes number of triangles as well.

Hoikas: And done

PR #86, opened 2017-01-12
Python refactor 3

zrax:
The third and final part of the Python bindings refactor.  This one adds a few minor tweaks (some methods in `hsBounds3Ext`, a macro for checking deleted properties, rename string conversion helpers, etc), but the primary change is to unify the patterns for using vectors/arrays with Python:
* Now, everything that returns a non-map collection will use a Python tuple.
  * This protects against "silent" bugs like trying to append to or modify the temporary returned object, since tuples cannot be modified directly.
  * Ideally we would use a proxy to the underlying vector to make it more Pythonic, but that would be a much larger change, and would likely break backwards compatibility if it wasn't done carefully.
* Everything that accepts a variable-sized collection from Python can now accept any sequence object, using Python's PySequence_Fast API.  This means it uses fast access for lists and tuples, and only performs one conversion for other types of sequences.
  * In order to simplify this, a new wrapper object is created for creating and managing the lifetime of the PySequence_Fast returned object: `pySequenceFastRef`
  * Some fixed-size and all mixed-type receivers will still only allow a tuple, since those are handled differently and are not iterated over in the same manner.

### DIFF
--- a/Python/Module.cpp
+++ b/Python/Module.cpp
@@ -148,7 +148,7 @@ PY_METHOD_GLOBAL_VA(PyHSPlasma, CleanFileName,
         PyErr_SetString(PyExc_TypeError, "CleanFileName expects a string");
         return NULL;
     }
-    return PlStr_To_PyStr(CleanFileName(fname, allowPathChars != 0));
+    return pyPlasma_convert(CleanFileName(fname, allowPathChars != 0));
 }
 
 }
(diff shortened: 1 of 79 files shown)


Hoikas:
Something odd is going on with keyframe exporting in Korman. We're exporting a layer animation with BezScalarKeys. When setting the `keys` attribute on `plLeafController`, it complains that it wants point3 keyframes however, which is nonsense for opacity animations...

zrax:
Should be fixed now :trollface:

Hoikas:
This appears to export legal ages from Korman now. @Deledrius we need another TEST 😛

PR #87, opened 2017-01-13
Add CMake option to allow building without PhysX support.

Deledrius:
Allows building without linking to PhysX even on systems that have the SDK available and detected.


zrax:
I would suggest that if `ENABLE_PHYSX` is `OFF`, we also shouldn't bother running the `find_package` for it.

Deledrius:
Good point.  I was concerned with disabling it post-detection and forgot about that.  I'll update this shortly.

PR #90, opened 2017-01-18
Make better use of string_theory's stdio module in the tools

zrax:
This mostly converts the various `printf`s to `ST::printf` and removes `printf`s where unnecessary.  There are a few other cleanups/fixes too though.


Hoikas:
It makes sense on a read through...

PR #92, opened 2017-01-21
Linkage cleanup

zrax:
"C" linkage is only needed for symbols directly exported to or imported from a C API.  Symbols which are only accessible through C++ code and/or pointers don't need the "C" linkage, so this cleans up everything else.


Deledrius:
👍 Nice clean-up.  No issues compiling, and I can still import PyHSPlasma too.

PR #96, opened 2017-01-30
Fix missing return statements in pyCreatable conversion cases.

Deledrius:


Deledrius:
Returning the Creatable after conversion instead of just dropping it on the floor and falling through the remaining cases is probably the better option.

Hoikas:
:trollface:

Issue #97, opened 2017-02-01
Port to string_theory regressed opening some prp files


diafero:
Attached is an example of a prp file that used to open fine with older versions of libhsplasma, that works fine with Uru, but that does not work with current master. The error shown by PrpShop says something about invalid UTF-8, so I assume this is related to the recent move to a different string processing library.

[Elodea_District_beach.prp.tar.gz](https://github.com/H-uru/libhsplasma/files/745202/Elodea_District_beach.prp.tar.gz)

diafero:
The issue is probably caused by an object (a texture, more specifically) having a name that's not valid UTF-8: `Holz alt f�r Balk.004`

Hoikas:
I noticed that a different blend file produced by Dulcamara included textures with latin_1 encoded ü characters. This actually made Blender 2.7x crash! IMO the bug is on the end user's side for creating defective files, but I'll defer to @zrax.

Deledrius:
Calling the file 'defective' is a bit harsh.  That's an input problem on the part of Blender for not converting the encoding to what is needs/expects.

diafero:
Well, the point is, the file works just fine in Blender and Uru. My dataserver generation scripts load the PRP files via libhsplasma to make some sanity checks (e.g. to make sure there are no horribly broken plLinkResponders in there), and thus dataserver generation fails with libhsplasma master. The way things are right now, I have the choice between asking Dulcamara (and others using umlauts in filenames and object names) to rename some things in order to fix issues they can't see, or alternatively forever sticking to an older version of libhsplasma. I don't think the first first solution would be appropriate; after all, the files *are* working perfectly fine, so there's no fundamental reason anything needs to be changed. Many age writers are not very tech-savvy, and I don't even know how hard it is to rename things and fix every reference in Blender. Ultimately, the issue only exists on my side, in the toolchain I use, which (somewhat arbitrarily) places more restrictions on the validity of PRP files than Uru does. Even though I agree it would be nice to have everything be valid UTF-8, I feel it is wrong to put the burden of this decision onto age writers. It's not like we have too many of them...
(Considering the age of Uru:TPOTS, I would actually be surprised if it does use UTF-8, but I don't know.)

Deledrius:
@diafero, do your automated checks function properly with 65fdff591ea3 applied?

PR #98, opened 2017-02-02
Latin-1 PRP support

zrax:
This allows libhsplasma to load PRPs with Latin-1 encoded strings (see #97).

Note that doing so and then re-saving the PRP will cause the string to get converted to UTF-8, which **may break external references** if they are not also updated to UTF-8.


RalfJung:
Does this mean that if the file containing the external reference is also loaded and saved with libhsplasma, then everything will continue to work?

PR #100, opened 2017-06-14
Python Camera Bindings

Hoikas:
This changeset implements python bindings for every single doggone camera class I could find. Supercedes #80. Please note that some commits that were rebased from ancient history have been modified to fix issues and omissions.


Python/PRP/Message/pyCameraMsg.cpp
```diff
@@ -0,0 +1,166 @@
...
+PY_PROPERTY_PROXY(plCameraConfig, CameraMsg, config, getConfig)
+PY_PROPERTY(bool, CameraMsg, activated, getActivated, setActivated)
+
+static PyGetSetDef pyCameraMsg_GetSet[] = {
+    pyCameraMsg_newCam_getset,
+    pyCameraMsg_triggerer_getset,
+    pyCameraMsg_transTime_getset,
+    pyCameraMsg_config_getset,
+    pyCameraMsg_activated_getset
+};
```
Hoikas: :boom:

core/PRP/Animation/plLineFollowMod.h
```diff
@@ -41,7 +41,7 @@ class PLASMA_DLL plLineFollowMod : public plMultiModifier {
     };
 
 protected:
-    FollowMode fFollowMode;
+    unsigned short fFollowMode;
```
Hoikas: At the time of writing, I didn't really see the value in "strongly" typing this since this is a plain old enum and not an enum class. Wrapping the enum type into Python seemed like an excessive amount of work considering. I can go back and wrap the enum if you prefer, but... :tongue:

Hoikas:
Thanks for the detailed review @zrax. I will start revising this PR later today.

Python/PRP/Animation/pyLineFollowMod.cpp
```diff
@@ -0,0 +1,145 @@
...
+    Py_RETURN_NONE;
+}
+
+PY_METHOD_VA(LineFollowMod, delStereizer, "Params: idx") {
+    Py_ssize_t idx;
+    if (!PyArg_ParseTuple(args, "n", &idx)) {
+        PyErr_SetString(PyExc_TypeError, "delStereizer expects an int");
+        return NULL;
+    }
+    if (idx < self->fThis->getStereizers().size()) {
```
Hoikas: I looked at the other cases where this check is done--they all cast idx to `size_t` and don't check for negative values. Is that acceptable or should I change my type in `PyArg_ParseTuple`?

zrax: Following the existing pattern is probably good enough.  Casting it to an unsigned type will result in a value that cannot be less than zero, and the range check will work (even if a bit funny) :)

Hoikas: :trollface:

Hoikas:
Fixed issues from review.

Python/Module.cpp
```diff
@@ -781,6 +797,8 @@ PyMODINIT_FUNC initPyHSPlasma() {
     PyModule_AddObject(module, "plConvexIsect", Init_pyConvexIsect_Type());
     PyModule_AddObject(module, "plSoundMsg", Init_pySoundMsg_Type());
     PyModule_AddObject(module, "plSwimMsg", Init_pySwimMsg_Type());
+    PyModule_AddObject(module, "plCameraMsg", Init_pyCameraMsg_Type());
+    PyModule_AddObject(module, "plAnimPath", Init_pyAnimPath_Type());
```
Hoikas: The reason why these are listed above the last two types in this section is because they are in the wrong part of the list. Future work will be to move them into the correct section.

Hoikas:
Bah. one sneaky bastard snuck through.

Hoikas:
NOW all issues have been fixed from review. I love the conflicts that github is reporting. That's gonna be **really** hard to merge... *not*. 🙄

Hoikas:
Rebased to fix the green merge button

PR #104, opened 2017-11-27
Implement pyFogEnvironment

dpogue:


Python/Module.cpp
```diff
@@ -801,6 +802,7 @@ PyMODINIT_FUNC initPyHSPlasma() {
     PyModule_AddObject(module, "plAnimPath", Init_pyAnimPath_Type());
     PyModule_AddObject(module, "plAvLadderMod", Init_pyAvLadderMod_Type());
     PyModule_AddObject(module, "plLadderModifier", Init_pyLadderModifier_Type());
+    PyModule_AddObject(module, "plFogEnvironment", Init_pyFogEnvironment_Type());
```
Hoikas: Should be with the main keyed object types list above (side note: the ladder types are in the wrong places)

PR #105, opened 2017-12-03
Correctly encapsulate UruSP key controllers

Hoikas:
Incorrect plController export reported by @dpogue https://github.com/H-uru/korman/issues/78#issuecomment-336614840

### DIFF
--- a/core/PRP/Animation/plLeafController.cpp
+++ b/core/PRP/Animation/plLeafController.cpp
@@ -16,6 +16,9 @@
 
 #include "plLeafController.h"
 #include "plKeyControllers.hpp"
+#include "plPosController.h"
+#include "plRotController.h"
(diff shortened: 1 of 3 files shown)


core/PRP/Animation/plLeafController.cpp
```diff
@@ -437,6 +440,29 @@ plLeafController* plLeafController::CompactToLeafController() const {
     return ctrl;
 }
 
+plController* plLeafController::EncapsulateKeyController() const {
+    plLeafController* expanded = ExpandToKeyController();
```
zrax: This leaks in all but the default case

zrax: No it doesn't -- ignore me :/

Hoikas:
:trollface:

PR #110, opened 2017-12-27
Subworld Bindings

Hoikas:
This also includes the ability to export plHKSubworld to CWE. See H-uru/Plasma#555


Python/PRP/Message/pySubWorldMsg.cpp
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#include "pySimulationMsg.h"
+
+#include "PRP/KeyedObject/pyKey.h"
+#include <PRP/Message/plSimulationMsg.h>
+
+PY_PLASMA_NEW(SubWorldMsg, plSubWorldMsg)
+
+PY_PROPERTY(plKey, SubWorldMsg, subworld, getWorldKey, setWorldKey);
```
zrax: Shouldn't this property be `worldKey`?

Hoikas: Fixed

PR #112, opened 2018-04-04
Fix unusable pyMipmaps from DDSurfaces.

Deledrius:
DDSurface::createMipmap returns raw plMipmaps with uninitialized KeyedObjects, leaving them partially unusable in PyHSPlasma scripts.  This adds the KeyedObject init to the function's Python binding.

### DIFF
--- a/Python/Util/pyDDSurface.cpp
+++ b/Python/Util/pyDDSurface.cpp
@@ -77,10 +77,19 @@ PY_METHOD_VA(DDSurface, setFromMipmap,
     Py_RETURN_NONE;
 }
 
-PY_METHOD_NOARGS(DDSurface, createMipmap,
+PY_METHOD_VA(DDSurface, createMipmap,
+    "Params: Mipmap name\n"
     "Create a new plMipmap from this plDDSurface")
 {
-    return ICreate(self->fThis->createMipmap());
+    const char* name = "";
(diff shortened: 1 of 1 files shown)


Python/Util/pyDDSurface.cpp
```diff
@@ -77,10 +77,19 @@ PY_METHOD_VA(DDSurface, setFromMipmap,
...
 }
 
-PY_METHOD_NOARGS(DDSurface, createMipmap,
+PY_METHOD_VA(DDSurface, createMipmap,
+    "Params: Mipmap name"
     "Create a new plMipmap from this plDDSurface")
 {
-    return ICreate(self->fThis->createMipmap());
+    const char* name;
+    if (!PyArg_ParseTuple(args, "s", &name)) {
```
zrax: I feel like this should be optional, in order to a) not break existing code, and b) more closely match the general `hsKeyedObject` constructor.  The key should be `init`ed either way, but with an empty string if no parameter is passed (again, to match `hsKeyedObject`).

Deledrius: I didn't think that a blank key name would be valid (it causes silent failures when I did that during testing, as only the first one is kept).  That's still a better solution than no KO at all, as you can easily set the key name from Python.

PR #117, opened 2018-07-07
Add Appveyor Config

zrax:


appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+      -DCMAKE_INSTALL_PREFIX="%APPVEYOR_BUILD_FOLDER%/dist/libhsplasma"
+      -DENABLE_PYTHON=ON -DENABLE_NET=ON -DENABLE_TOOLS=ON -DENABLE_PHYSX=OFF
+      %CMAKE_PARAMS% ..
+  - cmake --build . --config Release
+  - cmake --build . --config Release --target INSTALL
+  - cd "%APPVEYOR_BUILD_FOLDER%\dist"
+  - cmake -E tar cf libhsplasma.zip --format=zip libhsplasma
+
+artifacts:
+  - path: dist\libhsplasma.zip
```
Deledrius: It might be useful to have the artifact name contain the architecture and Python version.

Deledrius:
I tested a simple import of each PyHSPlasma in the appropriate versions of Python, followed by a simple `dir(PyHSPlasma)`.  Both 3.6 libraries worked fine, but the Python 2.7 one throws a uselessly vague IOError:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  IOError: [Errno 0] Error
```
This could be an error in my Python 2.7 configuration, so it would be good to have someone else test and confirm or dismiss this error.

Deledrius:
Looks good.  The only remaining question I have is whether there's a reason you chose to execute everything in the `build` step, instead of using the appropriate `install` step for the dependencies, `before_build` for the project generator, the `after_build` step for the install, etc.?

zrax:
What advantage does using those steps offer?  I wasn't sure if the environment was persisted across all the steps, so I just put them together for the sake of simplicity (read: laziness).

Deledrius:
Yes, the environment persists; the entire purpose is to allow you to set up the environment and perform steps that build on each other.  I don't think there's an explicit benefit in this particular case, but it does make the discrete steps more clearly labeled, and the logs will report where the failure was too.  I don't think it's essential, but it would be more clear.  We can fix this later, however.

PR #122, opened 2018-07-27
Implement stable key partial sorting for dependent key lists

### LINKED ISSUE #121: SceneObject modifier list race conditions
It is possible to export data that will not function due to race conditions in all versions of Uru. Specifically, `plAGMasterMod` assumes that all other modifiers are loaded on the targeted `plSceneObject` and expects to be able to access the `plAGModifier` immediately. Given that the `plSceneObject` modifier list is often what causes these objects to load, there is a mandatory order for modifiers in the `plSceneObject` modifier list.

```
(arbitary mods)
plAGModifier
(arbitrary mods)
plAGMasterMod, plCameraModifier
```

libHSPlasma should probably correct this at write-time.

zrax:
Fixes #121 (at least for the currently known cases).

### DIFF
--- a/core/PRP/KeyedObject/plKey.cpp
+++ b/core/PRP/KeyedObject/plKey.cpp
@@ -149,21 +149,25 @@ std::vector<plKey> hsOrderKeys(const std::vector<plKey>& keys) {
     // std::sort cannot be used here, since hsKeyedObject::orderAfter does
     // not provide strict weak ordering, which is required for std::sort
     // to work correctly.  Instead, we perform a stable topological sort.
+    // TODO: This algorithm could probably be more efficient.
     std::vector<plKey> result;
     result.reserve(keys.size());
     std::list<plKey> input = std::list<plKey>(keys.begin(), keys.end());
 
-    while (input.size() > 1) {
+    while (!input.empty()) {
         auto lowest = input.begin();
         auto iter = lowest;
-        while (++iter != input.end()) {
-            if (lowest->orderAfter(*iter))
+        while (iter != input.end()) {
+            if (lowest->orderAfter(*iter)) {
                 lowest = iter;
+                iter = input.begin();
+            } else {
+                ++iter;
+            }
         }
         result.push_back(*lowest);
         input.erase(lowest);
     }
-    result.push_back(input.front());
 
     return result;
 }


core/PRP/Camera/plCameraModifier.cpp
```diff
@@ -155,6 +155,15 @@ void plCameraModifier::write(hsStream* S, plResManager* mgr) {
     S->writeBool(fResetAnimOnPop);
 }
 
+bool plCameraModifier::orderAfter(const hsKeyedObject* other) const {
+    // No strong association between plCameraModifier and plAGMasterMod,
+    // so just ensure this is always sorted after all plAGMasterMods
+    // TODO: Improve this if possible
+    if (other->getKey()->getType() == kAGMasterMod)
```
Hoikas: AGMasterMod is indeed the direct dependency (in some situations). However, CameraModifier absolutely must be after the AGModifier as well

core/PRP/Avatar/plAGMasterMod.cpp
```diff
@@ -74,6 +74,15 @@ void plAGMasterMod::write(hsStream* S, plResManager* mgr) {
     }
 }
 
+bool plAGMasterMod::orderAfter(const hsKeyedObject* other) const {
+    // This should be ordered after the plAGModifier of the same name
+    if (other->getKey()->getType() == kAGModifier
+            && other->getKey()->getName() == getKey()->getName()) {
```
Hoikas: Considering that this is run over the entire list of modifiers, the name comparison could likely be dropped.

zrax: The name comparison was intended to only create a strong dependency between associated keys.  For example, I assume it's perfectly fine for `[plAGModifier("cAnim01"), plAGMasterMod("cAnim01"), plAGModifier("cAnim01"), plAGMasterModifier("cAnim02")]` to remain interleaved, rather than sorting all AGModifiers before all AGMasterMods.

core/PRP/Camera/plCameraModifier.cpp
zrax: Since the AGMasterMod will cause all AGModifiers to be before their associated AGMasterMods, sorting the CameraModifier after all AGMasterMods will automatically ensure that CameraModifier is also after all AGModifiers.

core/PRP/Avatar/plAGMasterMod.cpp
Hoikas: It is, I'm just hoping there aren't any cases where the AGMaster and AGMod don't have the same key name. That's extremely unlikely IMO

zrax: My concern here was more about not reordering things that don't need to be reordered (i.e. clean round-tripping of existing, working PRP files).  Is the association stored somewhere else that it can be referenced for key sorting?

Hoikas: I highly doubt it, unfortunately

PR #123, opened 2018-08-12
Fix plAnimPath zero init

Hoikas:
Fixes random heap garbage being written to the disk, unless you feel like being explicit.

### DIFF
--- a/core/PRP/Animation/plAnimPath.h
+++ b/core/PRP/Animation/plAnimPath.h
@@ -45,7 +45,7 @@ protected:
    hsAffineParts fParts;
 
 public:
-    plAnimPath() : fController(NULL), fTMController(NULL) { }
+    plAnimPath() : fAnimPathFlags(), fMinDistSq(), fLength(), fController(), fTMController() { }
     virtual ~plAnimPath();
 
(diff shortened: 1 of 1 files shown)


core/PRP/Animation/plAnimPath.h
```diff
@@ -45,7 +45,7 @@ class PLASMA_DLL plAnimPath : public plCreatable {
    hsAffineParts fParts;
 
 public:
-    plAnimPath() : fController(NULL), fTMController(NULL) { }
+    plAnimPath() : fController(), fTMController(), fAnimPathFlags(), fMinDistSq(), fLength() { }
```
zrax: These should be reordered to match declaration order (else GCC warnings)

Hoikas: 😨  what's GCC smokin'?

Hoikas:
Updated to work around GCC's substance dependency issues.

zrax:
It may seem like GCC is just being picky, but keep in mind that the compiler will call the member initializers in the order they're declared in the class, _not_ in the order they're listed in the constructor.  So it does make some sense for the occasional case where that difference matters...

PR #127, opened 2018-08-20
Fix plJPEG image channels order

Hoikas:
Plasma's plMipmaps store image data in BGR(A) for all uncompressed image types. However, JPEGs are stored as RGB on disk, so we must flip the channels for images that we read in. RLE images are unaffected because they operate on the pixel level. Indeed, the previous behavior would be inconsistent in that Cyan's lossy JPEG mipmaps would appear in RGB while Cyan's RLE mipmaps would appear in BGR.

### DIFF
--- a/core/Util/plJPEG.cpp
+++ b/core/Util/plJPEG.cpp
@@ -192,6 +192,15 @@ void plJPEG::DecompressJPEG(hsStream* S, void* buf, size_t size)
         offs += out_stride;
     }
 
+    // Data stored as RGB on disk but Plasma uses BGR
+    uint32_t* dp = reinterpret_cast<uint32_t*>(buf);
+    for (size_t i=0; i<size; i += 4) {
+        *dp = (*dp & 0xFF00FF00)
+            | (*dp & 0x00FF0000) >> 16
+            | (*dp & 0x000000FF) << 16;
+        dp++;
+    }
+
     jpeg_finish_decompress(&ji.dinfo);
 }
 
--- a/core/Util/plJPEG.cpp
+++ b/core/Util/plJPEG.cpp
@@ -193,6 +193,9 @@ void plJPEG::DecompressJPEG(hsStream* S, void* buf, size_t size)
     }
 
     // Data stored as RGB on disk but Plasma uses BGR
+    if (reinterpret_cast<uintptr_t>(buf) % sizeof(uint32_t) != 0)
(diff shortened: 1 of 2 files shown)


zrax:
This one probably needs updates in PlasmaShop as well, for JPEG rendering and also JPEG decal alpha masking.

Hoikas:
That should be handled by H-uru/PlasmaShop#44

core/Util/plJPEG.cpp
```diff
@@ -189,6 +189,15 @@ void plJPEG::DecompressJPEG(hsStream* S, void* buf, size_t size) {
         offs += out_stride;
     }
 
+    // Data stored as RGB on disk but Plasma uses BGR
+    uint32_t* dp = reinterpret_cast<uint32_t*>(buf);
```
zrax, line 193: As in the PlasmaShop PR, we should probably add an assert to ensure the `buf` is properly aligned for the cast...

Deledrius:
> We've had to explain why certain mipmaps have the blue/red channels reversed when viewed in PRPShop on the OU discord. Perhaps it is time to revisit this and [H-uru/PlasmaShop#44](https://github.com/H-uru/PlasmaShop/pull/44), and @Deledrius's successor [H-uru/PlasmaShop#69](https://github.com/H-uru/PlasmaShop/pull/69)?

Yeah, this fix has been ready for years now. :(

core/Util/plJPEG.cpp
Hoikas: libHSPlasma doesn't seem to have a standard assertion macro like Plasma does, so I added a `throw hsBadParamException` if the alignment doesn't smell right.

zrax: Yeah, that's reasonable...

core/Util/plJPEG.cpp
```diff
@@ -192,6 +192,18 @@ void plJPEG::DecompressJPEG(hsStream* S, void* buf, size_t size)
         offs += out_stride;
     }
 
+    // Data stored as RGB on disk but Plasma uses BGR
+    if (reinterpret_cast<uintptr_t>(buf) % sizeof(uint32_t) != 0)
```
zrax, line 196:
```suggestion
    if (reinterpret_cast<uintptr_t>(buf) % alignof(uint32_t) != 0)
```

zrax (approved):

zrax:
Ugh, MSVC 2013 doesn't support `alignof()` apparently

PR #128, opened 2018-08-28
Fix improper re-use of PNG structs.

Deledrius:
Reading and writing of PNGs without actively using freshly-minted structs was causing bad data reads and writes.  This fix moves all PNG library structs local, to be discarded immediately after use.


core/Util/plPNG.h
```diff
@@ -30,23 +30,17 @@ class PLASMA_DLL hsPNGException : public hsException {
...
 public:
     static void DecompressPNG(hsStream* S, void* buf, size_t size);
     static void CompressPNG(hsStream* S, const void* buf, size_t size,
                             uint32_t width, uint32_t height, int pixelSize);
 
 private:
-    plPNG() : fPngReader(), fPngWriter(), fPngInfo(), fEndInfo() { }
-    ~plPNG();
+    plPNG() { }
+    ~plPNG() { }
```
Hoikas: This can be removed

core/Util/plPNG.h
```diff
@@ -30,23 +30,17 @@ class PLASMA_DLL hsPNGException : public hsException {
...
 
 public:
     static void DecompressPNG(hsStream* S, void* buf, size_t size);
     static void CompressPNG(hsStream* S, const void* buf, size_t size,
                             uint32_t width, uint32_t height, int pixelSize);
 
 private:
-    plPNG() : fPngReader(), fPngWriter(), fPngInfo(), fEndInfo() { }
-    ~plPNG();
+    plPNG() { }
```
Hoikas: If plPNG is no longer instantiated, perhaps the constructor could be marked deleted

core/Util/plPNG.cpp
```diff
@@ -50,96 +50,55 @@ plPNG& plPNG::Instance() {
...
 
-    pi.fEndInfo = png_create_info_struct(pi.fPngReader);
-    if (!pi.fEndInfo) {
-        png_destroy_read_struct(&pi.fPngReader, &pi.fPngInfo, NULL);
-        pi.fPngReader = NULL;
-        pi.fPngInfo = NULL;
+    pngInfo = png_create_info_struct(pngReader);
+    if (!pngInfo) {
+        png_destroy_read_struct(&pngReader, NULL, NULL);
+        pngReader = NULL;
```
Hoikas: No need to null a variable that's about to go out of scope.

core/Util/plPNG.cpp
```diff
@@ -170,22 +129,37 @@ void plPNG::DecompressPNG(hsStream* S, void* buf, size_t size) {
             throw hsPNGException(__FILE__, __LINE__, "PNG output buffer is too small");
     }
 
-    png_read_image(pi.fPngReader, rows.get());
-    png_read_end(pi.fPngReader, pi.fEndInfo);
+    png_read_image(pngReader, rows.get());
+    png_read_end(pngReader, endInfo);
+    
+    png_destroy_read_struct(&pngReader, &pngInfo, &endInfo);
```
Hoikas: There are a heck of a lot of lines dedicated to cleaning up this strict. I would consider using an `std::unique_ptr` to manage it. I did this with window handles in plUruLauncher

Deledrius:
Oh, whoops!  Forgot those.

zrax:
I was under the impression that `png_destroy_read_struct` and `png_destroy_write_struct` also freed the passed info structs...  The documentation implies that in any case.

Deledrius:
I was *just* checking that, and the documentation does explicitly say this.

PR #129, opened 2018-09-09
Add bindings for plSeekPointMod.

Deledrius:


Python/Module.cpp
```diff
@@ -612,6 +613,7 @@ PyMODINIT_FUNC initPyHSPlasma() {
     PyModule_AddObject(module, "plAGModifier", Init_pyAGModifier_Type());
     PyModule_AddObject(module, "plSingleModifier", Init_pySingleModifier_Type());
     PyModule_AddObject(module, "plAGMasterMod", Init_pyAGMasterMod_Type());
+    PyModule_AddObject(module, "plSeekPointMod", Init_pySeekPointMod_Type());
```
zrax: This must be initialized after plMultiModifier:

    >>> from PyHSPlasma import *
    ERROR: Base PyHSPlasma.plMultiModifier for type PyHSPlasma.plSeekPointMod is not initialized
    Classes derived from PyHSPlasma.plMultiModifier WILL NOT WORK CORRECTLY

Deledrius: Updated, and tested to actually import correctly.

PR #132, opened 2018-12-10
pyMipmap improvements

Hoikas:
This changeset adds more exception translators to the `pyMipmap` member functions and removes some redundant buffer copying.

### DIFF
--- a/Python/PRP/Surface/pyMipmap.cpp
+++ b/Python/PRP/Surface/pyMipmap.cpp
@@ -231,10 +231,9 @@ PY_METHOD_NOARGS(Mipmap, extractColorData,
     "Extract an RGB color buffer from a JPEG mipmap")
 {
     size_t dataSize = self->fThis->getWidth() * self->fThis->getHeight() * 3;
-    char* data = new char[dataSize];
+    PyObject* buf = PyBytes_FromStringAndSize(NULL, dataSize);
+    char* data = PyBytes_AS_STRING(buf);
     self->fThis->extractColorData(data, dataSize);
-    PyObject* buf = PyBytes_FromStringAndSize(data, dataSize);
-    delete[] data;
     return buf;
 }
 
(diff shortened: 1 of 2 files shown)


Python/PRP/Surface/pyMipmap.cpp
```diff
@@ -142,7 +142,12 @@ PY_METHOD_VA(Mipmap, setLevel,
         PyErr_SetString(PyExc_TypeError, "setLevel expects int, binary string");
         return NULL;
     }
-    self->fThis->setLevelData(level, (const void*)data, dataSize);
+    try {
+        self->fThis->setLevelData(level, (const void*)data, dataSize);
+    } catch (hsBadParamException& ex) {
```
zrax: Should catch by `const` reference

Python/PRP/Surface/pyMipmap.cpp
```diff
@@ -216,21 +231,19 @@ PY_METHOD_NOARGS(Mipmap, extractColorData,
     "Extract an RGB color buffer from a JPEG mipmap")
 {
     size_t dataSize = self->fThis->getWidth() * self->fThis->getHeight() * 3;
-    char* data = new char[dataSize];
+    PyObject* buf = PyBytes_FromStringAndSize(NULL, dataSize);
+    char* data = PyBytes_AS_STRING(buf);
```
zrax: This either needs to be added to PyPlasma.h for Python 2.x, or we need to formally drop support for Python 2.6 (I'd prefer the latter, but it's a slightly larger and more unrelated change)

Python/PRP/Surface/pyMipmap.cpp
Hoikas: Heh, I had it this way... Then noticed the rest of the file was catching by non-`const` reference. Fixed.

Python/PRP/Surface/pyMipmap.cpp
Hoikas: Fixed

PR #134, opened 2018-12-29
Allow PRPs to be read/written from any stream

Hoikas:
This enables PRPs to be read in from or written out to any arbitrary stream. The most readily apparent use case is writing a PRP to an `hsRAMStream` (or some custom derivative of `hsStream`, if you're feeling really fancy). As a bonus, some temporary heap allocations were removed.

### DIFF
--- a/core/ResManager/plResManager.cpp
+++ b/core/ResManager/plResManager.cpp
@@ -204,8 +204,13 @@ plPageInfo* plResManager::ReadPagePrc(const pfPrcTag* root) {
 }
 
 void plResManager::WritePage(const ST::string& filename, plPageInfo* page) {
-    hsFileStream* S = new hsFileStream();
-    S->open(filename, fmWrite);
+    hsFileStream S;
+    S.open(filename, fmWrite);
+    WritePage(&S, page);
+    S.close();
+}
+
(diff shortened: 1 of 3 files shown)


core/ResManager/plResManager.cpp
```diff
@@ -156,33 +156,42 @@ plPageInfo* plResManager::ReadPage(const ST::string& filename, bool stub) {
...
-    hsFileStream* S = new hsFileStream();
-    if (!S->open(file, fmRead))
-        throw hsFileReadException(__FILE__, __LINE__, file.c_str());
+    if (packed) {
+        hsFileStream prpS;
+        if (!prpS.open(file, fmRead))
+            throw hsFileReadException(__FILE__, __LINE__, file.c_str());
+        // hsFileStream destructor closes the file
+        return ReadPage(&prpS, &prpS, stub);
+    } else {
```
zrax: Does anyone actually use this code path?  We could probably get rid of it to simplify things a bit...

Hoikas: @dpogue

dpogue: Realistically, nobody is going to use this code, but it allows for some fun experimenting so I'd prefer to keep it if it's not a maintenance burden (which it doesn't seem to be?)

PR #135, opened 2019-01-01
Inline exception classes to silence "class derived from non-dllexport base" warnings in MSVC 2017

zrax:
Also makes better use of ST literals in exception messages.


Deledrius:
Hooray!  Cleaner build logs!

PR #137, opened 2019-01-02
Allow plAgeInfo IO with streams

Hoikas:


core/ResManager/plAgeInfo.cpp
```diff
@@ -62,25 +67,28 @@ void plAgeInfo::readFromFile(const ST::string& filename) {
...
         else
             eType = plEncryptedStream::kEncXtea;
-        ((plEncryptedStream*)S)->open(filename, fmCreate, eType);
+        S.open(filename, fmCreate, eType);
+        writeToStream(&S, ver);
+        S.close();
     }
+}
 
+void plAgeInfo::writeToStream(hsStream* S, PlasmaVer) const {
```
zrax: The `PlasmaVer` parameter is unused here -- it should probably be removed.

Hoikas: Done

PR #138, opened 2019-01-14
Read and write the DisplayToWorld transform rotation quat symmetrically.

zrax:
plClient writes a computed DisplayToWorld transform, but does some additional computation during the read in preparation for setting up the Havok physical.  We may want this computed transform at some point, but for now it's easier (and better for preserving PRP identity) to just read and write the same bits avoiding all additional computation.


dpogue (approved):
:+1: LGTM

zrax:
Superseded by #251

PR #144, opened 2019-08-07
Make plKey a movable type and make better use of it as such.

zrax:


zrax:
It's actually for better temporary management.  Consider:

```
thing->setKey(returnsATemporaryKey())
```

If `setKey` takes a const reference, the key gets incremented twice and decremented once.  If we instead accept a key by value and move it, the temporary is automatically copy-elided, and we only increment its reference count once total.

Hoikas:
Cool! I didn't think about that situation.

PR #146, opened 2019-09-15
Add bindings for plSharedMesh.

Deledrius:


dpogue:
My router dies, and you folks have the most plProductive weekend in years
without me 🙄

Carry on 👍

Deledrius:
That's how I felt when my power was out for two days this week, @dpogue.  I *was* wondering where you were...

Hoikas:
I guess that explains where all my gremlins escaped to...

PR #149, opened 2019-09-27
PyHSPlasma SDL

Hoikas:
This adds just enough SDL to PyHSPlasma for one to be able to poke around the SDL files.

### DIFF
--- a/Python/SDL/pyStateDescriptor.cpp
+++ b/Python/SDL/pyStateDescriptor.cpp
@@ -1,0 +1,189 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * HSPlasma is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with HSPlasma.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "pySDL.h"
+
+#include <SDL/plStateDescriptor.h>
+
+PY_PLASMA_DEALLOC(StateDescriptor)
+PY_PLASMA_EMPTY_INIT(StateDescriptor)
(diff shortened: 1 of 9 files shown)


Python/SDL/pyStateDescriptor.cpp
```diff
@@ -0,0 +1,183 @@
...
+        return 0;
+    }
+
+    PyErr_SetString(PyExc_TypeError, "StateDescriptor subscript should be an int or a string");
+    return -1;
+}
+
+PY_PLASMA_REPR_DECL(StateDescriptor)
+{
+    return PyString_FromSTString(self->fThis->getName());
```
zrax: The `repr` string should include more than just the name of the descriptor...  Maybe something like `<plStateDescriptor {name}>`?

Python/SDL/pyStateDescriptor.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+PY_METHOD_VA(StateDescriptor, delVariable,
+             "Params: variable\n"
+             "Deletes the requested variable by name or index")
+{
+    Py_ssize_t idx;
+    if (PyArg_ParseTuple(args, "n", &idx)) {
+        try {
+            self->fThis->delVariable(idx);
+        } catch (...) {
```
zrax: Same comments as above

Python/SDL/pyVarDescriptor.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include <SDL/plStateDescriptor.h>
+
+PY_PLASMA_DEALLOC(VarDescriptor)
+PY_PLASMA_EMPTY_INIT(VarDescriptor)
+PY_PLASMA_NEW(VarDescriptor, plVarDescriptor)
+
+PY_PLASMA_REPR_DECL(VarDescriptor)
+{
+    return PyString_FromSTString(self->fThis->getName());
```
zrax: Similar to above, maybe `<plVarDescriptor {name}>`?

Python/SDL/pyStateDescriptor.cpp
Hoikas: This is actually the implementation of the `__str__()` method, which shares its signature with `__repr__()`. I can introduce a proper `__repr__()` if you prefer.

zrax: Sneaky; I didn't catch that you were assigning it to `tp_str` :trollface:.  In that case, I might prefer a new macro (even if it's almost identical to `PY_PLASMA_REPR_DECL`) to make it clearer in the code.

Hoikas:
Updated to fix issues from review.

PR #150, opened 2019-09-28
Support `os.PathLike` for paths and filenames.

### LINKED ISSUE #148: PyHSPlasma should accept `os.PathLike`
PyHSPlasma functions that accept an `str` for a filename or path should also accept an object implementing `os.PathLike` by calling the object's `__fspath__` method. `os.PathLike` was added in Python 3.6.

Hoikas:
This changeset adds support for the `os.PathLike` protocol added in Python 3.6 for any properties or function arguments that expect anything resembling a path or a filename. This allows us to pass `pathlib.Path` objects seemlessly to libHSPlasma in Python 3.6+. CPython seems to use `PyUnicode_FSDecoder` instead of manually calling the `__fspath__` method, so that is the approach I took. Unfortunately, In lower versioned Python 3.5 and lower, it is still required to convert path objects to their string representations before passing them to libHSPlasma.

This closes #148

### DIFF
--- a/Python/PyPlasma.cpp
+++ b/Python/PyPlasma.cpp
@@ -44,6 +44,25 @@ ST::string PyAnyString_AsSTString(PyObject* str)
     }
 }
 
+int PyAnyString_PathDecoder(PyObject* obj, void* str)
+{
+#if (PY_MAJOR_VERSION > 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION >= 2))
+    PyObject* fsConvert;
+    if (PyUnicode_FSDecoder(obj, &fsConvert)) {
+        *((ST::string*)str) = PyAnyString_AsSTString(fsConvert);
+        Py_DECREF(fsConvert);
+        return 1;
+    }
+    return 0;
+#else
+    if (PyAnyString_Check(obj)) {
+        *((ST::string*)str) = PyAnyString_AsSTString(obj);
+        return 1;
+    }
+    return 0;
+#endif
+}
+
 int PyType_CheckAndReady(PyTypeObject* type)
 {
     static std::unordered_set<PyTypeObject*> init_bases;
--- a/Python/PyPlasma.h
+++ b/Python/PyPlasma.h
@@ -26,7 +26,8 @@
 PyObject* PyString_FromSTString(const ST::string& str);
 PyObject* PyUnicode_FromSTString(const ST::string& str);
 ST::string PyAnyString_AsSTString(PyObject* str);
-#define PyAnyString_Check(ob) (PyUnicode_Check(value) || PyBytes_Check(value))
+int PyAnyString_PathDecoder(PyObject* obj, void* str);
+#define PyAnyString_Check(ob) (PyUnicode_Check(ob) || PyBytes_Check(ob))
 
 int PyType_CheckAndReady(PyTypeObject* type);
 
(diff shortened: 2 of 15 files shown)


Python/PyPlasma.cpp
```diff
@@ -60,3 +60,23 @@ int PyType_CheckAndReady(PyTypeObject* type)
...
+        *((PyObject**)result) = value;
+        Py_INCREF(value);
+
+        // Py_CLEANUP_SUPPORTED was not added until Python 3.1, so we will rely on the caller
+        // to clean up the reference.
+        return 1;
+    }
+
+    return 0;
+}
```
zrax: Rather than try to understand the compatibility differences between this and the versions in Python 3.2+ (hint, it has changed a fair bit even since 3.2), why not provide a new converter which does the conversion directly to `ST::string` and handles `os.PathLike` appropriately for Python 3.6+?

Hoikas: That seems fair. Would you prefer calling the `__fspath__()` method or continuing to use `PyUnicode_FSDecoder()`? The benefit to the latter is the ability to use `bytes` paths... I'm not sure why that is desirable, but it does give us that option.

zrax: I'm fine with using `PyUnicode_FSDecoder` when it's available.

Hoikas:
Ok, I've updated this to use a converter to `ST::string`. In doing so, I found and fixed an amusing bug in `PyAnyString_Check`. Also, included `pySDLMgr.readDescriptors`. Further, I updated a few of the type error messages. When there were a lot of arguments, I was uncertain how to clearly indicate what was going on, so thoughts there would be appreciated :smile:

Python/PyPlasma.cpp
```diff
@@ -44,6 +44,25 @@ ST::string PyAnyString_AsSTString(PyObject* str)
     }
 }
 
+int PyAnyString_PathDecoder(PyObject* obj, ST::string* str)
```
zrax: I think the second argument will need to be `void*` to comply with Python's API...

Python/PyPlasma.cpp
zrax: This function is no longer needed with your updates.

Python/PyPlasma.h
```diff
@@ -26,7 +26,8 @@
 PyObject* PyString_FromSTString(const ST::string& str);
 PyObject* PyUnicode_FromSTString(const ST::string& str);
 ST::string PyAnyString_AsSTString(PyObject* str);
-#define PyAnyString_Check(ob) (PyUnicode_Check(value) || PyBytes_Check(value))
+int PyAnyString_PathDecoder(PyObject* obj, ST::string* str);
+#define PyAnyString_Check(ob) (PyUnicode_Check(ob) || PyBytes_Check(ob))
```
zrax: Heh, good catch

Python/PyPlasma.cpp
Hoikas: It compiles and works on my testing in Windows. That's probably because Python's doing crazy va arg magic where type safety becomes a figment of the imagination. I can change it if you prefer, though.

Python/PyPlasma.h
Hoikas: This was one of those, "wait, what's going on here..." moments 😄

Python/PyPlasma.cpp
Hoikas: I went ahead and changed the second argument to `void*`.

zrax: I think newer versions of GCC (GCC 8?) will complain about it (our CI is only building with GCC 5, so it doesn't trigger).  So thanks for updating it.

PR #154, opened 2019-10-02
Switch to GitHub Actions for Linux/macOS CI builds

zrax:
This runs Ubuntu with GCC and Python 2.7 and 3.6, and macOS with clang and Python 3.7 (the Travis version tested macOS with Python 2.7).

Passing checks can be seen on:
https://github.com/zrax/libhsplasma/runs/245173590
https://github.com/zrax/libhsplasma/runs/245173542
https://github.com/zrax/libhsplasma/runs/245173559


Deledrius:
Would it be possible to do a build on macOS with GCC as well?

zrax:
@Deledrius, what would be the use case for testing/using GCC on macOS?

Deledrius:
Per discussions with @Hoikas, we think this may be the solution to a working build of the PyHSPlasma library compatible with Blender (which uses GCC), required for Korman.

PR #156, opened 2019-10-02
Fix crash in `pyEncryptedStream.IsFileEncrypted`.

Hoikas:
This fixes a crash encountered by Doobes when Korman thought it was a good idea to use `plEncryptedStream.IsFileEncrypted` on a directory path.

### DIFF
--- a/Python/Stream/pyEncryptedStream.cpp
+++ b/Python/Stream/pyEncryptedStream.cpp
@@ -38,8 +38,8 @@ PY_METHOD_VA(EncryptedStream, open,
             }
             Py_INCREF(self);
             return (PyObject*)self;
-        } catch (...) {
-            PyErr_SetString(PyExc_IOError, "Error opening file");
+        } catch (const std::exception& ex) {
+            PyErr_SetString(PyExc_IOError, ex.what());
             return nullptr;
         }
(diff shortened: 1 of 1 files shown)


Python/Stream/pyEncryptedStream.cpp
```diff
@@ -112,7 +112,12 @@ PY_METHOD_STATIC_VA(EncryptedStream, IsFileEncrypted,
         PyErr_SetString(PyExc_TypeError, "IsFileEncrypted expects a string or an os.PathLike object");
         return nullptr;
     }
-    return pyPlasma_convert(plEncryptedStream::IsFileEncrypted(filename));
+    try {
+        return pyPlasma_convert(plEncryptedStream::IsFileEncrypted(filename));
+    } catch (...) {
+        PyErr_SetString(PyExc_IOError, "Error opening file");
```
zrax: If you catch `const std::exception&`, you can get the actual failure from libhsplasma from `std::exception::what()`.

Hoikas: I copy pasta'd this from all the other error handlers in the file ^_^

Hoikas:
Updated to fix all pyEncryptedStream exception handlers to use `std::exception`.

PR #161, opened 2019-10-20
Restore compatibility with MSVC 2013.

### LINKED ISSUE #160: Visual C++2013 build broken.
`hsArraySize` uses features not available in VC++2013. There also seem to be lots of errors around the union in the ENDSWAP functions. Visual C++2013 is required by Korman due to Blender 2.79b's dependency on the same. It is presently not feasible to upgrade Korman to Blender 2.80.

zrax:
Fixes #160.


Deledrius:
Builds for me!

PR #162, opened 2019-10-27
Expose Vault to PyHSPlasma

Hoikas:
As best I could tell, `plVaultStore` owns the nodes and generally speaking returns copies, so all the nodes types are value-wrapped.

### DIFF
--- a/Python/Vault/pyVaultNode.h
+++ b/Python/Vault/pyVaultNode.h
@@ -1,0 +1,95 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * HSPlasma is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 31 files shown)


Python/Vault/pyVaultNode.h
```diff
@@ -0,0 +1,92 @@
...
+PY_WRAP_PLASMA_VALUE(VaultChronicleNode, plVaultChronicleNode);
+PY_WRAP_PLASMA_VALUE(VaultPlayerInfoListNode, plVaultPlayerInfoListNode);
+PY_WRAP_PLASMA_VALUE(VaultMarkerNode, plVaultMarkerNode);
+PY_WRAP_PLASMA_VALUE(VaultAgeInfoNode, plVaultAgeInfoNode);
+PY_WRAP_PLASMA_VALUE(VaultAgeInfoListNode, plVaultAgeInfoListNode);
+PY_WRAP_PLASMA_VALUE(VaultMarkerListNode, plVaultMarkerListNode);
+
+inline PyObject* pyPlasma_convert(const plVaultNode& node)
+{
+    if (plVaultPlayerNode* access = node.upcastToPlayerNode())
```
zrax: Each of these upcast calls checks fNodeType independently, making this somewhat inefficient...  Why not use a switch instead?

Hoikas: Good point. This was mostly me being lazy, not wanting to have to bang out the cast myself, due to the scope of the changes. I'll go back and hit this.

Hoikas: Updated to fix this

Hoikas:
:boom:

PR #164, opened 2019-11-14
Expose PRC getter in PyHSPlasma Creatables.

Deledrius:


Python/PRP/pyCreatable.cpp
```diff
@@ -104,6 +105,33 @@ PY_METHOD_VA(Creatable, write,
     Py_RETURN_NONE;
 }
 
+PY_GETSET_GETTER_DECL(Creatable, prc)
+{
+    hsRAMStream* S = new hsRAMStream();
+    pfPrcHelper* prc = new pfPrcHelper(S);
```
Hoikas: These could be stack allocated.

Python/PRP/pyCreatable.cpp
```diff
@@ -104,6 +105,33 @@ PY_METHOD_VA(Creatable, write,
...
+    S->read(dataSize, buf);
+
+    auto data = ST::string::from_utf8((const char*)buf, dataSize);
+    PyObject* result = PyString_FromSTString(data);
+    delete[] buf;
+
+    return result;
+}
+
+PY_PROPERTY_SETTER_MSG(Creatable, prc, "To set the contents, you must re-create the object")
```
Hoikas: Forgot to update the error text, eh? ;)

Python/PRP/pyCreatable.cpp
```diff
@@ -104,6 +105,33 @@ PY_METHOD_VA(Creatable, write,
...
+    pfPrcHelper* prc = new pfPrcHelper(S);
+    self->fThis->prcWrite(prc);
+
+    S->rewind();
+    size_t dataSize = S->size() - S->pos();
+    unsigned char* buf = new unsigned char[dataSize];
+    S->read(dataSize, buf);
+
+    auto data = ST::string::from_utf8((const char*)buf, dataSize);
+    PyObject* result = PyString_FromSTString(data);
```
Hoikas: You could use something like `PyString_FromStringAndSize` to prevent copying into a temporary `ST::string`.

Python/PRP/pyCreatable.cpp
Deledrius: I did update it...

Python/PRP/pyCreatable.cpp
Deledrius: Nifty!

Python/PRP/pyCreatable.cpp
zrax: Why is this a property?  Creatables have no such member or property -- if anything, this could be a function on plCreatable which is then exposed to PyHSPlasma (also as a function).

Deledrius: It's a property because I had originally intended it to function as one, as an interface to prcParse and prcWrite.  I never got around to implementing the setter to cover prcParse, and a script @Hoikas and I are using required my existing getter, so I submitted this as-is with the setter essentially stubbed out.

If it's preferable (due to a lack of interest in setting the object PRC from Python), I could rewrite this as an ordinary function.

core/PRP/plCreatable.cpp
```diff
@@ -46,6 +47,15 @@ void plCreatable::prcParse(const pfPrcTag* tag, plResManager* mgr)
     }
 }
 
+ST::string plCreatable::toPrc(pfPrcHelper::PrcExclude exclude)
+{
+    hsRAMStream S;
+    pfPrcHelper prc(&S);
+    prc.exclude(exclude);
+    prcWrite(&prc);
+    return ST::string::from_utf8((const char*)S.data(), S.size(), ST::utf_validation_t::assume_valid);
```
zrax: FYI, you can shorten the last parameter to `ST::assume_valid`

PR #166, opened 2019-12-02
Python 3.8 compatibility

zrax:
~~Plus a fix for CMake 3.16~~


zrax:
I removed the CMake changes, since they break more than they fix, and my local CMake issue was fixed by just deleting the cache and regenerating.

PR #169, opened 2019-12-10
Suggested fix for issue #168 Python CMakeLists incorrectly determines…

### LINKED ISSUE #168: Python CMakeLists incorrectly determines PYTHON_LIB_DIR
The current `Python/CMakeLists.txt` is responsible for setting the destination directory of the shared python library `PyHSPlasma.so`.  It currently does so by executing a Python code snippet obtaining the value of `distutils.sysconfig.get_python_lib()`.

Unfortunately it does so with a call to that routine with no arguments, which sets `PYTHON_LIB_DIR` to the installation directory of the currently running Python interpreter.

While that may be appropriate for building/installing as a manner of installing HSPlasma for the machine performing the build, it becomes inappropriate when  HSPlasma is being built in a build environment such as Jenkins, where the Jenkins build process attempts to do a `make install` with `CMAKE_INSTALL_PREFIX` set to a non-live destination.  If the build is performed (as it should be) in  Jenkins under a non-privileged user, the `make install` operation will attempt to install `PyHSPlasma.so` into the working Python library directories, and fail because of privileges.  As it should.  If for no other reason than that `make install` ignored `CMAKE_INSTALL_PREFIX`.

I am suggesting for the `NOT WIN32` situation to alter your snippet to use

  `get_python_lib(prefix='${CMAKE_INSTALL_PREFIX}')`

instead of the base `get_python_lib()` call.  This will properly pick up the destination for the Python library in an isolated build environment.

-rarified

tallpgr:
… PYTHON_LIB_DIR

Allow ${CMAKE_INSTALL_PREFIX} to override Python's installation prefix
when determining where to place PyHSPlasma.so library for Python.

Fixes #168.

### DIFF
--- a/Python/CMakeLists.txt
+++ b/Python/CMakeLists.txt
@@ -752,7 +752,7 @@ if(NOT WIN32)
                           SUFFIX ".so"
     )
 
-    execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import sys, distutils.sysconfig; sys.stdout.write(distutils.sysconfig.get_python_lib())"
+    execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import sys, distutils.sysconfig; sys.stdout.write(distutils.sysconfig.get_python_lib(prefix='${CMAKE_INSTALL_PREFIX}'))"
                     OUTPUT_VARIABLE _PYTHON_LIB_DIR
     )
     install(TARGETS PyHSPlasma


Hoikas:
It appears that the commit is not associated with your github account. You'll want to correct your git config for this repository to use the email address associated with your github account, recommit, and repush.

Also, you may want to include the exact string `Fixes #168` so that the corresponding issue will be ensured to close on merge of this PR. You've included some very close variants of that throughout, but I am uncertain if github will pick up those variants.

tallpgr:
I have updated GitHub to include the email address in the pull request and commit.  If you are sure the commit message requires “Fixes #168” I’ll do another commit on this branch.  But if the title needs Fixes #168 I’ll do another PR.

Hoikas:
Adding `Fixes #168` to either the title/contents of the pull request or the commit message will work :smile:. It's your option as to where to put it.

Deledrius:
> But if the title needs Fixes #168 I’ll do another PR.

No need for a new PR, just edit this one. :)

zrax:
Ideally we would have an option to allow overriding the install directory for the Python module...  But this is still definitely an improvement over always assuming the system location.

PR #170, opened 2019-12-18
Pack DSpan GeoSpans into single vertex buffer.

Hoikas:
This changeset optimizes the output of `plDrawableSpans::composeGeometry` by attempting to pack spans into a one vertex buffer in plGBufferGroup. These vertex buffers correspond to DirectX vertex buffers in Plasma. There are many buffer group formats in Plasma, and the code attempts to not obliterate anything another user of the buffer group has done.

This also fixes an issue in which the index data of a `plGeometrySpan` could be lost if the number of indices changed after being initialized or set by IO.

### DIFF
--- a/core/PRP/Geometry/plDrawableSpans.cpp
+++ b/core/PRP/Geometry/plDrawableSpans.cpp
@@ -661,14 +661,12 @@ void plDrawableSpans::clearSpans()
     fSpanSourceIndices.clear();
 }
 
-size_t plDrawableSpans::addIcicle(const plIcicle& span)
+size_t plDrawableSpans::addIcicle(plIcicle* span)
 {
-    plIcicle* iceCopy = new plIcicle(span);
-    fIcicles.push_back(iceCopy);
-    size_t iceId = fIcicles.size() - 1;
-    fSpans.push_back(fIcicles[iceId]);
-    fSpanSourceIndices.push_back(iceId | kSpanTypeIcicle);
-    return fSpans.size() - 1;
+    fIcicles.push_back(span);
+    fSpans.push_back(span);
+    fSpanSourceIndices.push_back((fIcicles.size() - 1) | kSpanTypeIcicle);
+    return fIcicles.size() - 1;
 }
 
 size_t plDrawableSpans::createBufferGroup(unsigned char format)
@@ -744,9 +742,22 @@ void plDrawableSpans::setSpaceTree(plSpaceTree* tree)
     fSpaceTree = tree;
 }
 
+size_t plDrawableSpans::IFindBufferGroup(unsigned int format, unsigned int vertsNeeded)
+{
+    // For some reason, Plasma attempts to maintain only one vertex storage buffer per group
+    for (size_t i = 0; i < fGroups.size(); ++i) {
+        plGBufferGroup* group = fGroups[i];
+        if (group->getFormat() == format) {
+            if (group->getNumVertices() + vertsNeeded < plGBufferGroup::kMaxVertsPerBuffer)
+                return i;
+        }
+    }
+
+    return createBufferGroup(format);
+}
+
 void plDrawableSpans::composeGeometry(bool clearspans, bool calcbounds)
 {
-    std::map<unsigned int, std::pair<plGBufferGroup*, size_t> > groups;
     for (auto group = fGroups.begin(); group != fGroups.end(); ++group)
         delete *group;
     for (auto span = fSpans.begin(); span != fSpans.end(); ++span)
@@ -754,44 +765,7 @@ void plDrawableSpans::composeGeometry(bool clearspans, bool calcbounds)
 
     for (size_t i=0; i<fSourceSpans.size(); i++) {
         plGeometrySpan* span = fSourceSpans[i];
-        unsigned int format = span->getFormat();
-        plGBufferGroup* group = groups[format].first;
-        if (!group) {
-            group = new plGBufferGroup(format);
-            groups[format] = std::make_pair(group, fGroups.size());
-            fGroups.push_back(group);
-        }
-        size_t buf_idx = group->getNumVertBuffers();
-        std::vector<unsigned short> indices = span->getIndices();
-        std::vector<plGeometrySpan::TempVertex> verts = span->getVertices();
-
-        plGBufferCell cell;
-        cell.fColorStart = -1;
-        cell.fVtxStart = 0;
-        cell.fLength = verts.size();
-        std::vector<plGBufferCell> cell_tmp;
-        cell_tmp.push_back(cell);
-
-        std::vector<plGBufferVertex> new_verts;
-        new_verts.resize(verts.size());
-        for (size_t j=0; j<verts.size(); j++) {
-            plGeometrySpan::TempVertex v1 = verts[j];
-            plGBufferVertex v2;
-            v2.fPos = v1.fPosition;
-            v2.fNormal = v1.fNormal;
-            for (size_t k=0; k<8; k++) {
-                v2.fUVWs[k] = v1.fUVs[k];
-            }
-            for (size_t k=0; k<3; k++) {
-                v2.fSkinWeights[k] = v1.fWeights[k];
-            }
-            v2.fSkinIdx = v1.fIndices;
-            v2.fColor = v1.fColor;
-            new_verts[j] = v2;
-        }
-        group->addIndices(indices);
-        group->addVertices(new_verts);
-        group->addCells(cell_tmp);
+        size_t group = IFindBufferGroup(span->getFormat(), span->getNumVertices());
 
         auto mat_f = std::find(fMaterials.begin(), fMaterials.end(), span->getMaterial());
         size_t mat_idx = -1;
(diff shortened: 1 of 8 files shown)


core/PRP/Geometry/plDrawableSpans.cpp
```diff
@@ -661,13 +661,11 @@ void plDrawableSpans::clearSpans()
...
+size_t plDrawableSpans::addIcicle(plIcicle* span)
 {
-    plIcicle* iceCopy = new plIcicle(span);
-    fIcicles.push_back(iceCopy);
-    size_t iceId = fIcicles.size() - 1;
-    fSpans.push_back(fIcicles[iceId]);
-    fSpanSourceIndices.push_back(iceId | kSpanTypeIcicle);
+    fIcicles.push_back(span);
+    fSpans.push_back(span);
+    fSpanSourceIndices.push_back((fSpans.size() - 1) | kSpanTypeIcicle);
```
zrax: I assume this change to use the span index instead of the icicle index is intentional?

core/PRP/Geometry/plGBufferGroup.h
```diff
@@ -90,6 +93,12 @@ class PLASMA_DLL plGBufferGroup
         kStoreIsDirty = 0x4
     };
 
+    enum
+    {
+        kMaxNumVertsPerBuffer = 32000,
+        kMaxNumIndicesPerBuffer = 32000,
```
zrax: I feel like "Num" is redundant here...

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,113 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+    ice->setIStartIdx(idxOffset);
+    ice->setILength(geoSpan->getNumIndices());
+}
+
+void plGBufferGroup::IPackVertexSpan(const plGeometrySpan* geoSpan, plVertexSpan* vSpan)
+{
+    size_t i;
+    for (i = 0; i < fVertBuffStorage.size(); ++i)
+        if (geoSpan->getNumVertices() + getNumVertices(i) < kMaxNumVertsPerBuffer)
+            break;
```
zrax: Please use braces around multiline loop bodies, even if they are not syntactically necessary.

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,113 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
         fIdxBuffStorage[idx][i] = indices[i];
 }
 
+void plGBufferGroup::IPackIcicle(const plGeometrySpan* geoSpan, plIcicle* ice)
+{
+    size_t i;
+    for (i = 0; i < fIdxBuffStorage.size(); ++i)
+        if (geoSpan->getNumIndices() + fIdxBuffCounts[i] < kMaxNumIndicesPerBuffer)
+            break;
```
zrax: Please use braces around multiline loop bodies, even if they are not syntactically necessary.

core/PRP/Geometry/plDrawableSpans.cpp
```diff
@@ -744,54 +742,29 @@ void plDrawableSpans::setSpaceTree(plSpaceTree* tree)
...
 }
 
+size_t plDrawableSpans::IFindBufferGroup(unsigned int format, unsigned int vertsNeeded)
+{
+    // For some reason, Plasma attempts to maintain only one vertex storage buffer per group
+    for (size_t i = 0; i < fGroups.size(); ++i) {
+        plGBufferGroup* group = fGroups[i];
+        if (group->getFormat() == format)
+            if (group->getNumVertices() + vertsNeeded < plGBufferGroup::kMaxNumVertsPerBuffer)
+                return i;
```
zrax: Please use braces around multiline if bodies, even if they are not syntactically necessary.

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,113 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+        memcpy(vertStorage, fVertBuffStorage[i], vertOffset);
+    fVertBuffSizes[i] = vertBuffSize;
+    delete fVertBuffStorage[i];
+    fVertBuffStorage[i] = vertStorage;
+
+    float* destp = (float*)(vertStorage + vertOffset);
+    for (const auto& vertex : geoSpan->getVertices()) {
+        *destp++ = vertex.fPosition.X;
+        *destp++ = vertex.fPosition.Y;
+        *destp++ = vertex.fPosition.Z;
```
zrax: This type of type punning can generate very slow unaligned writes (yes, I know we already do that in other places :( )...  A memcpy might be better:

```c++
unsigned char *destp = vertStorage + vertOffset;
...
memcpy(destp, vertex.fPosition, sizeof(vertex.fPosition));
destp += sizeof(vertex.fPosition);
```

core/PRP/Geometry/plGBufferGroup.h
Hoikas: Probably so, but I copy-pasta'd this directly from Plasma. While we at it, maybe the values should be changed to INT16_MAX?

core/PRP/Geometry/plDrawableSpans.cpp
Hoikas: I don't remember having a specific intention there. I'll change it since in libHSPlasma fSpans should effectively be a copy of fIcicles.

core/PRP/Geometry/plGBufferGroup.cpp
Hoikas: Good point. Perhaps a pass over the old code would be relevant at some point.

Hoikas: I was a little spooked by copying a class directly into POD, so I added a static_assert on the vector data being the expected size.

core/PRP/Geometry/plDrawableSpans.cpp
```diff
@@ -661,14 +661,12 @@ void plDrawableSpans::clearSpans()
...
-    plIcicle* iceCopy = new plIcicle(span);
-    fIcicles.push_back(iceCopy);
-    size_t iceId = fIcicles.size() - 1;
-    fSpans.push_back(fIcicles[iceId]);
-    fSpanSourceIndices.push_back(iceId | kSpanTypeIcicle);
-    return fSpans.size() - 1;
+    fIcicles.push_back(span);
+    fSpans.push_back(span);
+    fSpanSourceIndices.push_back((fIcicles.size() - 1) | kSpanTypeIcicle);
+    return fIcicles.size() - 1;
```
zrax: And now you've changed the other part (the return value)...  Hmm, I really don't remember how all this works, nor why we keep two separate vectors.  Who is using these values anyway?

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,115 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+                *((uint32_t*)destp) = vertex.fIndices; destp++;
+            }
+        }
+
+        static_assert(sizeof(vertex.fNormal) == sizeof(float) * 3, "vertex normal unexpected size");
+        memcpy(destp, &vertex.fNormal, sizeof(vertex.fNormal));
+        destp += sizeof(vertex.fNormal);
+
+        *((uint32_t*)destp) = vertex.fColor; destp++;
+        *((uint32_t*)destp) = vertex.fSpecularColor; destp++;
```
zrax: These should be converted to memcpy as well.

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,115 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+        static_assert(sizeof(vertex.fPosition) == sizeof(float) * 3, "vertex position unexpected size");
+        memcpy(destp, &vertex.fPosition, sizeof(vertex.fPosition));
+        destp += sizeof(vertex.fPosition);
+
+        int weightCount = (fFormat & kSkinWeightMask) >> 4;
+        if (weightCount > 0) {
+            for (size_t j = 0; j < weightCount; ++j)
+                *destp++ = vertex.fWeights[j];
+            if (fFormat & kSkinIndices) {
+                *((uint32_t*)destp) = vertex.fIndices; destp++;
```
zrax: Memcpy

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,115 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+    unsigned int vertOffset = fVertBuffSizes[i];
+    unsigned int vertBuffSize = vertOffset + (geoSpan->getNumVertices() * fStride);
+    unsigned char* vertStorage = new unsigned char[vertBuffSize];
+    if (vertOffset != 0)
+        memcpy(vertStorage, fVertBuffStorage[i], vertOffset);
+    fVertBuffSizes[i] = vertBuffSize;
+    delete fVertBuffStorage[i];
+    fVertBuffStorage[i] = vertStorage;
+
+    float* destp = (float*)(vertStorage + vertOffset);
```
zrax: This needs to be a `char` sized pointer for `destp += sizeof(thing)` to work correctly.

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,115 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+    for (const auto& vertex : geoSpan->getVertices()) {
+        static_assert(sizeof(vertex.fPosition) == sizeof(float) * 3, "vertex position unexpected size");
+        memcpy(destp, &vertex.fPosition, sizeof(vertex.fPosition));
+        destp += sizeof(vertex.fPosition);
+
+        int weightCount = (fFormat & kSkinWeightMask) >> 4;
+        if (weightCount > 0) {
+            for (size_t j = 0; j < weightCount; ++j)
+                *destp++ = vertex.fWeights[j];
```
zrax: This can probably be a single memcpy without the loop...

core/PRP/Geometry/plGBufferGroup.cpp
```diff
@@ -547,6 +560,115 @@ void plGBufferGroup::addIndices(const std::vector<unsigned short>& indices)
...
+        destp += sizeof(vertex.fNormal);
+
+        *((uint32_t*)destp) = vertex.fColor; destp++;
+        *((uint32_t*)destp) = vertex.fSpecularColor; destp++;
+
+        for (size_t j = 0; j < (fFormat & kUVCountMask); ++j) {
+            static_assert(sizeof(vertex.fUVs[j]) == sizeof(float) * 3, "vertex UV unexpected size");
+            memcpy(destp, &vertex.fUVs[j], sizeof(vertex.fUVs[j]));
+            destp += sizeof(vertex.fUVs[j]);
+        }
```
zrax: This can probably be a single memcpy without the loop...

core/PRP/Geometry/plGBufferGroup.cpp
zrax: Yep, the `static_assert`s are definitely reasonable here...

core/PRP/Geometry/plDrawableSpans.cpp
Hoikas: I suspect the reason there are two vectors is because we match what Plasma does. It used to have many different kinds of not necessarily icicle spans. It's all mostly dead end now though... For the sake of IO, we really only care about the icicle spans.

Hoikas:
Force-pushed fixes for review

PR #173, opened 2020-01-01
Return pyKeyFrames as the correct subtype

dpogue:
Returning the `.keys` of a plLeafController in Python would return a tuple containing an array of hsKeyFrames. In reality, that should be an array of the specific hsKeyFrame subclasses.

/cc @Jrius This should resolve the need for haxKeyFrames in ZLZ

### DIFF
--- a/Python/PRP/Animation/pyKeys.h
+++ b/Python/PRP/Animation/pyKeys.h
@@ -17,17 +17,48 @@
 #ifndef _PYKEYS_H
 #define _PYKEYS_H
 
+#include <PRP/Animation/hsKeys.h>
+
 #include "PyPlasma.h"
 
-PY_WRAP_PLASMA(KeyFrame, struct hsKeyFrame);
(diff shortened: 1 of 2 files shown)


Python/PRP/Animation/pyKeys.h
```diff
@@ -17,17 +17,48 @@
 #ifndef _PYKEYS_H
 #define _PYKEYS_H
 
+#include <PRP/Animation/hsKeys.h>
```
Hoikas: Why not use forward declarations?

dpogue: That wouldn't allow me to call `key->getType()` in the switch statement

Jrius:
> /cc @Jrius This should resolve the need for haxKeyFrames in ZLZ
I'm always glad to remove hack code ! Will fix this next time I work on ZLZ. Thanks !

PR #174, opened 2020-02-01
Add pyAgeLinkEffects bindings.

Hoikas:


Python/PRP/Message/pyLinkToAgeMsg.cpp
```diff
@@ -46,8 +44,31 @@ PY_GETSET_SETTER_DECL(LinkToAgeMsg, ageLink)
...
+{
+    // This cannot be a subclass, since it's an inline member
+    return pyAgeLinkEffects_FromAgeLinkEffects(&self->fThis->getLinkEffects());
+}
+
+PY_GETSET_SETTER_DECL(LinkToAgeMsg, linkEffects)
+{
+    PY_PROPERTY_CHECK_NULL(linkEffects)
+
+    plAgeLinkEffects effects = self->fThis->getLinkEffects();
```
zrax: This needs to be a reference in order for the assignment below to work correctly...

PR #177, opened 2020-02-16
Support MacOS newlines in hsStream

Jrius:
Some files (only `spyroom.age` AFAIK) use MacOS `\r` line endings, which cause `hsStream::readLine()` to fail as it assumes those are followed by a `\n` character. This fixes that.

### DIFF
--- a/core/Stream/hsStream.cpp
+++ b/core/Stream/hsStream.cpp
@@ -193,8 +193,17 @@ ST::string hsStream::readLine()
     }
     if (c != '\n' && c != '\r')
         line.append_char(c);
-    if (c == '\r')
-        readByte(); // Eat the \n in Windows-style EOLs
+    if (c == '\r') {
+        if (!eof()) {
+            // Eat the \n in Windows-style EOLs
+            char nextChar = readByte();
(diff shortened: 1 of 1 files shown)


Hoikas:
Looks overall fine to me, but I'll defer to @zrax.

core/Stream/hsStream.cpp
```diff
@@ -193,8 +193,17 @@ ST::string hsStream::readLine()
...
-    if (c == '\r')
-        readByte(); // Eat the \n in Windows-style EOLs
+    if (c == '\r') {
+        if (!eof()) {
+            // Eat the \n in Windows-style EOLs
+            char nextChar = readByte();
+            if (nextChar != '\n') {
+                // Old-school MacOS carriage return without line feed...
+                // Rewind one character.
+                skip(-1);
```
zrax: Hmm, ideally we would use a peek here, but it doesn't look like we ever implemented a peek() for hsStream :/.  This is particularly bad for encrypted streams, which start over from the beginning of the file on a negative skip, so we'll probably want to come back and update this some day.

PR #181, opened 2020-06-16
Add plConvexIsect::SinglePlane python bindings

Jrius:
Needed to add support for softvolumes in ZLZ, figured it would be a good time to finally get familiar with writing Python bindings...

Allows accessing the list of planes of a `plConvexIsect`, and retrieve each plane's position, normal, etc.

This is heavily inspired by the bindings for the very similar `plCameraModifier::CamTrans` class. It required making `SinglePlane`'s fields protected and add getter/setter methods, and move some of the PRC read/write logic from `plConvexIsect` into `SinglePlane`.

### DIFF
--- a/Python/PRP/Region/pyConvexIsect.cpp
+++ b/Python/PRP/Region/pyConvexIsect.cpp
@@ -63,7 +63,7 @@ PY_GETSET_GETTER_DECL(ConvexIsect, planes)
 {
     PyObject* list = PyTuple_New(self->fThis->getPlanes().size());
     for (size_t i = 0; i < self->fThis->getPlanes().size(); i++)
-        PyTuple_SET_ITEM(list, i, pySinglePlane_FromSinglePlane(&(self->fThis->getPlanes()[i])));
+        PyTuple_SET_ITEM(list, i, pySinglePlane_FromSinglePlane(self->fThis->getPlanes()[i]));
     return list;
 }
 
@@ -98,7 +98,9 @@ PY_PLASMA_IFC_METHODS(ConvexIsect, plConvexIsect)
 
 /* pySinglePlane */
 
-PY_PLASMA_NEW(SinglePlane, plConvexIsect::SinglePlane)
+PY_PLASMA_VALUE_DEALLOC(SinglePlane)
+PY_PLASMA_EMPTY_INIT(SinglePlane)
+PY_PLASMA_VALUE_NEW(SinglePlane, plConvexIsect::SinglePlane)
 
 PY_PROPERTY(hsVector3, SinglePlane, norm, getNorm, setNorm)
 PY_PROPERTY(hsVector3, SinglePlane, pos, getPos, setPos)
@@ -119,6 +121,8 @@ PY_PLASMA_TYPE(SinglePlane, plConvexIsect::SinglePlane, "plConvexIsect::SinglePl
 
 PY_PLASMA_TYPE_INIT(SinglePlane)
 {
+    pySinglePlane_Type.tp_dealloc = pySinglePlane_dealloc;
+    pySinglePlane_Type.tp_init = pySinglePlane___init__;
     pySinglePlane_Type.tp_new = pySinglePlane_new;
     pySinglePlane_Type.tp_getset = pySinglePlane_GetSet;
     if (PyType_CheckAndReady(&pySinglePlane_Type) < 0)
@@ -128,4 +132,4 @@ PY_PLASMA_TYPE_INIT(SinglePlane)
     return (PyObject*)&pySinglePlane_Type;
 }
 
-PY_PLASMA_IFC_METHODS(SinglePlane, plConvexIsect::SinglePlane)
+PY_PLASMA_VALUE_IFC_METHODS(SinglePlane, plConvexIsect::SinglePlane)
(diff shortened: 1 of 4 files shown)


Python/PRP/Region/pyVolumeIsect.h
```diff
@@ -18,8 +18,10 @@
 #define _PYVOLUMEISECT_H
 
 #include "PyPlasma.h"
+#include <PRP/Region/plVolumeIsect.h>
 
 PY_WRAP_PLASMA(VolumeIsect, class plVolumeIsect);
+PY_WRAP_PLASMA(SinglePlane, class plConvexIsect::SinglePlane);
 PY_WRAP_PLASMA(ConvexIsect, class plConvexIsect);
```
Hoikas, line 25: Since you have now included the declarations of `plVolumeIsect`, `plConvexIsect::SinglePlane`, and `plConvexIsect`, you may safely delete their forward declarations. In other words, remove the `class` from each of these lines.

core/PRP/Region/plVolumeIsect.cpp
```diff
@@ -158,49 +158,94 @@ void plConeIsect::IPrcParse(const pfPrcTag* tag, plResManager* mgr)
...
+    fDist = tag->getParam("Dist", "0").to_float();
+    fWorldDist = tag->getParam("WorldDist", "0").to_float();
+
+    const pfPrcTag* child = tag->getFirstChild();
+    while (child) {
+        if (child->getName() == "Normal") {
+            if (child->hasChildren())
+                fNorm.prcParse(child->getFirstChild());
+        }
+        else if (child->getName() == "Position") {
```
Hoikas, line 211: No line break between closing brace and else.

core/PRP/Region/plVolumeIsect.h
```diff
@@ -75,12 +75,32 @@ class PLASMA_DLL plConvexIsect : public plVolumeIsect
...
+    public:
+        hsVector3 getNorm() const { return fNorm; }
+        hsVector3 getPos() const { return fPos; }
+        hsVector3 getWorldNorm() const { return fWorldNorm; }
+        float getDist() const { return fDist; }
+        float getWorldDist() const { return fWorldDist; }
+
+        void setNorm(hsVector3 norm) { fNorm = norm; }
+        void setPos(hsVector3 pos) { fPos = pos; }
+        void setWorldNorm(hsVector3 worldNorm) { fWorldNorm = worldNorm; }
```
Hoikas, line 101: These setters make an extra copy of the `hsVector3` passed in. You can correct this by moving the temporary by eg `fNorm = std::move(norm);`

Python/PRP/Region/pyConvexIsect.cpp
```diff
@@ -57,18 +59,73 @@ PyMethodDef pyConvexIsect_Methods[] = {
...
 
+PY_GETSET_GETTER_DECL(ConvexIsect, planes)
+{
+    PyObject* list = PyTuple_New(self->fThis->getPlanes().size());
+    for (size_t i = 0; i < self->fThis->getPlanes().size(); i++)
+        PyTuple_SET_ITEM(list, i, pySinglePlane_FromSinglePlane(&(self->fThis->getPlanes()[i])));
+    return list;
+}
+
+PY_PROPERTY_SETTER_MSG(ConvexIsect, planes, "To add planes, use addPlane")
```
Hoikas, line 70: `addPlane` should probably be changed to accept a `plConvexIset::SinglePlane` if the expected arguments of position, normal are not supplied.

Python/PRP/Region/pyVolumeIsect.h
```diff
@@ -18,8 +18,10 @@
 #define _PYVOLUMEISECT_H
 
 #include "PyPlasma.h"
+#include <PRP/Region/plVolumeIsect.h>
 
 PY_WRAP_PLASMA(VolumeIsect, class plVolumeIsect);
+PY_WRAP_PLASMA(SinglePlane, class plConvexIsect::SinglePlane);
```
zrax, line 24: This probably makes more sense as a `PY_WRAP_PLASMA_VALUE`, using copying semantics rather than pointers and ownership...

Jrius:
Thanks for the feedback, will make those changes when I can !

Jrius:
This should cover most of your suggestions. `plConvexIsect::addPlane` is overloaded to support either a `SinglePlane` or two `hsVector3`, but only on the C++ side. On the Python side, I left only the version with two `hsVector3` because I wasn't sure how to write the proper documentation for it (I'm not sure it would be Pythonic at all). Let me know what you think.

Also, should I add `ClassName()`, `read()`, `write()`, `toPrc()` and `fromPrc()` methods to `SinglePlane`'s Python bindings ? This appears to be the norm for most of the classes in the bindings.

core/PRP/Region/plVolumeIsect.h
```diff
@@ -95,8 +115,12 @@ class PLASMA_DLL plConvexIsect : public plVolumeIsect
     void IPrcParse(const pfPrcTag* tag, plResManager* mgr) HS_OVERRIDE;
 
 public:
+    const std::vector<SinglePlane>& getPlanes() const { return fPlanes; }
+    std::vector<SinglePlane>& getPlanes() { return fPlanes; }
+
     /** Adds or updates a given plane */
     void addPlane(hsVector3 normal, const hsVector3& pos);
+    void addPlane(SinglePlane plane) { fPlanes.push_back(std::move(plane)); };
```
Hoikas, line 123: `emplace_back()`

Jrius:
Here we go. Sorry it took so long for such a small modification, RL stuff likes getting in the way...

Deledrius:
> 
> RL stuff likes getting in the way...

Certainly understandable.  Thank you for updating!

PR #183, opened 2020-07-24
Add plWarpMsg and pyWarpMsg.

Deledrius:
Adds plWarpMsg and associated Python bindings.

### DIFF
--- a/core/PRP/Message/plWarpMsg.h
+++ b/core/PRP/Message/plWarpMsg.h
@@ -1,0 +1,51 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * HSPlasma is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with HSPlasma.  If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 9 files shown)


core/PRP/Message/plWarpMsg.h
```diff
@@ -0,0 +1,51 @@
...
+    enum { kFlushTransform = 0x1, kZeroVelocity };
+
+protected:
+    uint32_t fWarpFlags;
+    hsMatrix44 fTransform;
+
+public:
+    plWarpMsg() : fWarpFlags(), fTransform() { }
+
+    uint32_t getWarpFlags() { return fWarpFlags; }
```
Hoikas, line 37:
```C++
uint32_t getWarpFlags() const { return fWarpFlags; }
```

core/PRP/Message/plWarpMsg.h
```diff
@@ -0,0 +1,51 @@
...
+    void setWarpFlags(uint32_t f) { fWarpFlags = f; }
+
+    hsMatrix44& getTransform() { return fTransform; }
```
Hoikas, line 40: Probably don't want to return a reference here.

core/PRP/Message/plWarpMsg.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "plMessage.h"
+#include "Math/hsMatrix44.h"
+
+class PLASMA_DLL plWarpMsg : public plMessage
+{
+    CREATABLE(plWarpMsg, kWarpMsg, plMessage)
+
+public:
+    enum { kFlushTransform = 0x1, kZeroVelocity };
```
zrax, line 28: Since these are (presumably ORed) flags, it makes sense to spell out `kZeroVelocity = 0x02` here

core/PRP/Message/plWarpMsg.h
```diff
@@ -0,0 +1,51 @@
...
+
+protected:
```
zrax, line 30: Variables with accessors and mutators should be private

PR #185, opened 2020-08-30
Fix round-tripping plPythonFileMod through HSPlasma.

Hoikas:
This fixes an issue in which PythonFileMods with string parameters would be shattered by round-tripping through HSPlasma. The spurious null terminator was being accumulated into the parameter, causing two nulls to be serialized. This caused a null byte to be passed to CWE's Python, resulting in a difficult to debug scenario where we had a seemingly valid string object in Python but the validation in `PyArg_ParseTuple` failed.

To reproduce this issue, load and save Neighborhood_District_nb01.prp in libHSPlasma, then visit the Age in CWE. The resulting error output will look [something like this](https://gist.github.com/Hoikas/4bc4e48b61a0ec89dee7a7ee72a37d57).

### DIFF
--- a/core/PRP/Modifier/plPythonFileMod.cpp
+++ b/core/PRP/Modifier/plPythonFileMod.cpp
@@ -82,10 +82,11 @@ void plPythonParameter::read(hsStream* S, plResManager* mgr)
     case kSubtitle:
         size = S->readInt();
         if (size == 0) {
-            fStrValue = "";
+            fStrValue = ST::null;
             return;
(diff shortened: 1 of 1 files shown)


zrax:
Did you check whether the null byte is there in all supported versions of Plasma?

Hoikas:
Good question - let me look at some EoA files.

Hoikas:
Yes, the strings in EoA also have the null terminator included in the length.

PR #187, opened 2020-10-01
Add a vcpkg triplet file and update README to describe its use.

zrax:
Also removes some obsolete info from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,8 @@ for all major operating systems.
 A typical build looks something like this:
 
     $ mkdir build
-    $ cd ./build
-    $ cmake ..
-
-To build with Python support run this cmake command:
-
-    $ cmake -DENABLE_PYTHON=ON ..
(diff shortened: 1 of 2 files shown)


Hoikas:
A good companion for this would be to add a python feature to the vcpkg libHSPlasma port - perhaps after Microsoft/vcpkg#13556 makes this easier. Note that using a static python would work fine on non-Windows platforms.

zrax:
> 
> 
> Note that using a static python would work fine on non-Windows platforms.

Yes, the readme section (and by extension, the vcpkg triplet file) are specific to Windows.  In my experience, other OSes work fine using their established dependency providers (Linux package managers, brew, etc).

libhsplasma doesn't require anything too crazy other than PhysX (for which we don't support the version in vcpkg yet anyway) for now :)

Hoikas (approved):

PR #189, opened 2020-10-04
Stub out read/write support for plWindBoneMod

dpogue:
Sadly even with knowing what the structure of the data is and knowing roughly how the values are used, it's still unclear what they actually represent :(

As far as I can tell, there is only a single example of a WindBoneMod in Myst V, used for the Laki'ahn trees.

### DIFF
--- a/core/PRP/Animation/plWindBoneMod.cpp
+++ b/core/PRP/Animation/plWindBoneMod.cpp
@@ -1,0 +1,113 @@
+/* This file is part of HSPlasma.
+ *
+ * HSPlasma is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


core/PRP/Animation/plWindBoneMod.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "plWindBoneMod.h"
+
+/* plWindBoneOscillator */
+void plWindBoneOscillator::read(hsStream* S)
+{
+    f00 = S->readFloat();
+    f04 = S->readFloat();
+    f08 = S->readFloat();
+    f0C = S->readFloat();
+    f10 = S->readFloat();
```
zrax, line 26: The usual convention in libhsplasma is to use names like `fUnknown1`, etc. which I think is easier to read...

zrax (approved):

PR #195, opened 2021-01-28
Main

sjaak-de-draak:
string_theory update (possibly breaking cmake for windows in the process)


sjaak-de-draak:
Thanks for setting me straight on the pull issue Zrax.
This pull request should be ignored and destroyed, right after I see what the appveyor makes of it ;)

PR #197, opened 2021-02-06
Add a vcpkg manifest.

Hoikas:
This adds a vcpkg manifest so there is less need to rely on the old PlasmaPrefix libs and to better enable adding/removing dependencies. I omitted `python3` from the manifest because vcpkg does not integrate with the deprecated `FindPythonLibs` and `FindPythonInterp` modules. I have also moved the "modern" Visual Studio compilers from AppVeyor to GHA to improve CI parallelism. The older compilers/pythons remain on AppVeyor.

Currently, there is still interest in Visual Studio 2013 and Python 3.5 because that is what Blender 2.79b is using on Windows.

~~Still TODO: switch AppVeyor over to using the vcpkg manifest.~~

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -1,0 +1,77 @@
+name: Windows-CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.platform.image }}
+    strategy:
+      matrix:
+        platform:
+          - { image: windows-2019, generator: Visual Studio 16 2019, cmake-arch: Win32, arch: x86, python: 3.9, str: windows-x86-v142 }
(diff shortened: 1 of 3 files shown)


Hoikas:
> Still TODO: switch AppVeyor over to using the vcpkg manifest.

I have found a bug in vcpkg.exe's ability to detect Visual Studio 2013 on AppVeyor, so this is going to block for awhile.

Hoikas:
Blocking on microsoft/vcpkg-tool#4.

Hoikas:
I'm backing off trying to use vcpkg with VS 2013 due to vcpkg's current baseline string_theory not supporting VS 2013. There's not really an easy way to override string_theory *only* when on the old compiler, so meh.

zrax (approved):

PR #199, opened 2021-02-10
Some more cmake package fixes.

Hoikas:
This fixes two errors with the generated cmake packages:
- `HSPlasmaNet-targets.cmake` was not being exported , leading to some fun linker errors down the road.
- Absolute library paths were being exported, making sharing the built library a challenging prospect. I fixed all cases of absolute paths except for when using the optional feature PhysX.

Bumps the minimum CMake requirement to 3.12. (Welcome to the real world.)


net/CMakeLists.txt
```diff
@@ -94,15 +94,11 @@ add_library(HSPlasmaNet
...
-endif()
-
-if(WIN32)
-    target_link_libraries(HSPlasmaNet crypt32)
-    target_link_libraries(HSPlasmaNet ws2_32)
-endif(WIN32)
+target_link_libraries(HSPlasmaNet
+                      HSPlasma string_theory
+                      $<$<NOT:$<PLATFORM_ID:SunOS>>:OpenSSL::Crypto>
+                      $<$<PLATFORM_ID:Windows>:ws2_32>
```
zrax, line 100: What happened to `crypt32`?  Is it not needed?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.12)
```
zrax, line 1: Huh, it looks like CI is using a newer CMake than what stock Ubuntu 18.04 ships...  Neat, I guess

net/CMakeLists.txt
Hoikas: Correct, `OpenSSL::Crypto` correctly links all of OpenSSL's transitive dependencies. This is a major benefit to using targets, where available 😄

CMakeLists.txt
Hoikas: Yeah, I think GHA runs the same CMake revisions across all of the hosted runners.

zrax (approved):

PR #200, opened 2021-02-25
The Revenge of Tsar Hoikas: PyHSPlasma Type Hints.

### LINKED ISSUE #190: PyHSPlasma Type Stubs
With the improvements in Python IDEs and Python language servers, such as VS Code, in both interpreting python type hints and displaying this information, we should consider writing [PEP 484 Type Stubs](https://www.python.org/dev/peps/pep-0484/#stub-files) for PyHSPlasma to improve the experience.

Hoikas:
I really wanted these while working on H-uru/korman#236 because I kept making trivial mistakes with misnaming attributes. These type hints were automatically generated using [`stubgen.py` from MyPy](https://github.com/python/mypy/blob/master/mypy/stubgen.py). I then manually curated them using the PyHSPlasma source code. When installed next to the .so/.pyd and using a supporting IDE, you get nice tooltips as shown below.

TypeHints

AFAICT, in general, .pyi files do not contain docstrings, so these were omitted. I'm not interested in adding them either - that's just too much work.

Closes #190.


dpogue:
Is it possible to generate this at build-time, rather than committing a frozen-in-time copy?

Hoikas:
No, see c5d386b.

Python/PyHSPlasma.pyi
```diff
@@ -0,0 +1,5616 @@
+# -*- coding: utf-8 -*-
+# copyage.py: Copy one age to another, with a new name and sequence prefix
+# Copyright (C) 2010 Branan Purvine-Riley
```
zrax, line 3: Not quite :trollface:

Python/PyHSPlasma.pyi
```diff
@@ -0,0 +1,5616 @@
...
+
+    def read(self, stream: hsStream) -> None: ...
+    def reset(self) -> None: ...
+    def write(self, stream: hsStream) -> None: ...
+
+class hsBitVector:
+    def clear(self) -> None: ...
+    def compact(self) -> None: ...
+    def getName(self, index : int) -> str: ...
+    def getValue(self, name : str) -> int: ...
```
zrax, line 248: Inconsistent whitespace

Python/PyHSPlasma.pyi
```diff
@@ -0,0 +1,5616 @@
...
+
+class plDebug:
+    kDLAll: int = ...
+    kDLDebug: int = ...
+    kDLError: int = ...
+    kDLNone: int = ...
+    kDLWarning: int = ...
+
+    @staticmethod
+    def Debug(msg : str) -> None: ...
```
zrax, line 1913: Inconsistent whitespace

Python/PyHSPlasma.pyi
```diff
@@ -0,0 +1,5616 @@
...
+    @staticmethod
+    def Error(msg : str) -> None: ...
```
zrax, line 1915: Inconsistent whitespace

Python/PyHSPlasma.pyi
```diff
@@ -0,0 +1,5616 @@
...
+    kLocalOnly: int = ...
+    kReserved: int = ...
+    kVolatile: int = ...
+
+    flags: int = ...
+    page: int = ...
+    prefix: int = ...
+    version: int = ...
+
+    def __init__(self, version : int = pvUnknown) -> None: ...
```
zrax, line 3773: Inconsistent whitespace

Python/PyHSPlasma.pyi
Hoikas: *Sigh* of course I copy from the file with this kind of nonsense...

zrax (approved):
I'd really like to find a way to make this more automated, but I guess it's Good Enough for Now (tm)

Hoikas:
Yeah, the challenge is that even if you use something like libtooling to generate the hints directly from the PyHSPlasma source code, there isn't enough information available to have the function arguments and return values at compile time. The only solution I can think of there is to encode the information into the docstring, but that might get burdensome for the overloaded methods. Some of the read-only properties also lack (Python) return types at compile time.

PR #201, opened 2021-04-15
Box bounds: generate indices for PotS/Havok

Jrius:
Found this bug when exporting Ages with Korman. Apparently, the PotS client will crash if box colliders don't have both vertices AND indices (strange, I know...)

### DIFF
--- a/core/PRP/Physics/plGenericPhysical.cpp
+++ b/core/PRP/Physics/plGenericPhysical.cpp
@@ -816,18 +816,6 @@ void plGenericPhysical::calcBoxBounds(size_t numPoints, const hsVector3* points)
     hsBounds3Corners corners = bounds.getCorners();
     fVerts.reserve(corners.size());
     std::copy(corners.begin(), corners.end(), std::back_inserter(fVerts));
-    fIndices.insert(fIndices.end(), {
-        0, 2, 1,
-        1, 2, 3,
-        0, 1, 4,
-        1, 5, 4,
-        0, 4, 2,
-        2, 4, 6,
-        1, 3, 7,
-        7, 5, 1,
-        3, 2, 7,
-        2, 6, 7,
-        4, 7, 6,
-        4, 5, 7
-    });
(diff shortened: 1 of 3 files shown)


Hoikas:
Oh, PotS... What would we do without you making our lives difficult?

core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -816,4 +816,18 @@ void plGenericPhysical::calcBoxBounds(size_t numPoints, const hsVector3* points)
...
+        0, 1, 4,
+        1, 5, 4,
+        0, 4, 2,
+        2, 4, 6,
+        1, 3, 7,
+        7, 5, 1,
+        3, 2, 7,
+        2, 6, 7,
+        4, 7, 6,
+        4, 5, 7
```
zrax, line 831: I might have a slight preference for putting this list in hsBounds3, since that's where the corners themselves are generated (and this relies on that generated ordering)...

zrax (approved):

core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -816,4 +816,6 @@ void plGenericPhysical::calcBoxBounds(size_t numPoints, const hsVector3* points)
     hsBounds3Corners corners = bounds.getCorners();
     fVerts.reserve(corners.size());
     std::copy(corners.begin(), corners.end(), std::back_inserter(fVerts));
+    fVerts.reserve(std::size(hsBounds3::CornerIndices));
```
zrax, line 819: `std::size` is C++17...  You could convert the `CornerIndices` to a `std::array` though to get around the need for it.

Jrius: Ah, that's what I was just wondering about. Thanks for the info. Give me a second...

Don't hesitate to nitpick if it's still suboptimal, my C++ isn't great.

zrax (approved):

PR #205, opened 2021-04-27
Pots boxphys offcentered vertices

Jrius:
PotS box bounds again. Generating the vertices/indices creates a box with face normals pointing inwards. Apparently doesn't bother PotS, but looks ugly once reimported in Blender. Fixed by reordering face indices.

Additionally, the algorithm to calculate corner vertices didn't support boxes that didn't encompass the world origin `(0, 0, 0)` due to a small initialization error, which resulted in one of the corner being placed at world origin. This fixes that too.

(If you want to see the wrongly generated vertices without this fix, paste the following code in a Python terminal. The generated box will be wrongly sized to encompass world origin.)
```
import PyHSPlasma as pl
gp = pl.plGenericPhysical()
gp.calcBoxBounds([pl.hsVector3(-1, -1, -1), pl.hsVector3(-.5,-.5, -.5)])
print(gp.verts)
```

### DIFF
--- a/core/PRP/Geometry/plDrawableSpans.cpp
+++ b/core/PRP/Geometry/plDrawableSpans.cpp
@@ -18,6 +18,7 @@
 #include "Debug/plDebug.h"
 #include "Util/hsRadixSort.h"
 #include <algorithm>
+#include <memory>
 
 /* plDISpanIndex */
 plDISpanIndex& plDISpanIndex::operator=(const plDISpanIndex& cpy) {
--- a/core/PRP/Geometry/plDrawableSpans.cpp
+++ b/core/PRP/Geometry/plDrawableSpans.cpp
@@ -528,10 +528,16 @@ void plDrawableSpans::calcBounds()
         hsBounds3Ext world;
 
         world.setFlags(hsBounds3Ext::kAxisAligned);
+        hsVector3* localPoints = new hsVector3[verts.size()];
+        hsVector3* worldPoints = new hsVector3[verts.size()];
         for (size_t j = 0; j < verts.size(); j++) {
-            loc += verts[j].fPos;
-            world += fIcicles[i]->getLocalToWorld().multPoint(verts[j].fPos);
+            localPoints[j] = verts[j].fPos;
+            worldPoints[j] = fIcicles[i]->getLocalToWorld().multPoint(verts[j].fPos);
         }
+        loc.setFromPoints(verts.size(), localPoints);
+        world.setFromPoints(verts.size(), worldPoints);
+        delete[] localPoints;
+        delete[] worldPoints;
         loc.unalign();
 
         fIcicles[i]->setLocalBounds(loc);
--- a/core/PRP/Geometry/plDrawableSpans.cpp
+++ b/core/PRP/Geometry/plDrawableSpans.cpp
@@ -528,16 +528,14 @@ void plDrawableSpans::calcBounds()
         hsBounds3Ext world;
 
         world.setFlags(hsBounds3Ext::kAxisAligned);
-        hsVector3* localPoints = new hsVector3[verts.size()];
-        hsVector3* worldPoints = new hsVector3[verts.size()];
+        auto localPoints = std::make_unique<hsVector3[]>(verts.size());
+        auto worldPoints = std::make_unique<hsVector3[]>(verts.size());
         for (size_t j = 0; j < verts.size(); j++) {
             localPoints[j] = verts[j].fPos;
             worldPoints[j] = fIcicles[i]->getLocalToWorld().multPoint(verts[j].fPos);
         }
-        loc.setFromPoints(verts.size(), localPoints);
-        world.setFromPoints(verts.size(), worldPoints);
-        delete[] localPoints;
-        delete[] worldPoints;
+        loc.setFromPoints(verts.size(), localPoints.get());
+        world.setFromPoints(verts.size(), worldPoints.get());
         loc.unalign();
 
         fIcicles[i]->setLocalBounds(loc);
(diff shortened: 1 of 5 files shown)


zrax (approved):

Hoikas:
@Jrius Would this issue also affect the bounds computed by `plDrawableSpans`, specifically the the `calcBounds()` method?

Jrius:
> 
> 
> @Jrius Would this issue also affect the bounds computed by `plDrawableSpans`, specifically the the `calcBounds()` method?

Hmm, looks like it, yeah. This might affect icicles and space trees... I'll have a closer look when I can, there might be other files with this issue.

Jrius:
Alright. I hunt down users of `operator+=`, and made them use the new `setFromPoints(unsigned int numPoints, const hsVector3* points)` instead. This fixes issues with physical box/sphere bounds and DSpan bounds/space trees. Also reduces the risk of reintroducing the issue in the future.

As usual, let me know if the syntax requires improvements. I'm not really fond of the `new[]..delete[]` in `plDrawableSpans::calcBounds()`, but it does the job.

I also added a new option to the ZLZ importer, which lets you import icicle bounds (local+world) and space trees from drawable spans. Should allow you to compare how Korman exports Ages with and without this fix ;)

Jrius:
Something else I noticed while playing with Korman+ZLZ. Objects with coordinate interfaces (for instance kickables and clickables) have their world bounds placed at the world origin instead of around the object itself. Space tree nodes also encompass those misplaced bounds. This doesn't happen in Cyan's Ages.

Not sure what causes this, but so far this hasn't been an issue for me - in my test Age those objects still display fine (they don't rely on realtime lighting though). Just mentioning in case it helps troubleshoot future bugs.

core/PRP/Geometry/plDrawableSpans.cpp
```diff
@@ -528,10 +528,16 @@ void plDrawableSpans::calcBounds()
         hsBounds3Ext world;
 
         world.setFlags(hsBounds3Ext::kAxisAligned);
+        hsVector3* localPoints = new hsVector3[verts.size()];
+        hsVector3* worldPoints = new hsVector3[verts.size()];
```
zrax, line 532: Instead of manual memory management, consider:

```c++
std::vector<hsVector3> localPoints(verts.size());
```

Or even:

```c++
auto localPoints = std::make_unique<hsVector3[]>(verts.size());
```

core/PRP/Region/hsBounds.cpp
```diff
@@ -176,6 +176,16 @@ hsBounds3Corners hsBounds3::getCorners() const
     return corners;
 }
 
+void hsBounds3::setFromPoints(unsigned int numPoints, const hsVector3* points)
```
zrax, line 179:
```suggestion
void hsBounds3::setFromPoints(size_t numPoints, const hsVector3* points)
```

core/PRP/Region/hsBounds.h
```diff
@@ -86,6 +86,7 @@ class PLASMA_DLL hsBounds3 : public hsBounds
     void setMaxs(const hsVector3& maxs) { fMaxs = maxs; }
     void setCenter(const hsVector3& center) { fCenter = center; }
 
+    void setFromPoints(unsigned int numPoints, const hsVector3* points);
```
zrax, line 89:
```suggestion
    void setFromPoints(size_t numPoints, const hsVector3* points);
```

core/PRP/Geometry/plDrawableSpans.cpp
Jrius: I swear, if only the courses I attended focused more on usage of the STD lib instead of constantly bashing it as "inefficient", I wouldn't have so many bad coding practices !

Doesn't build on Linux though, I'm looking into it...

Deledrius: Not sure about your experience, but when I was in school it was mostly true, to my recollection.  Simultaneously incomplete and unoptimized in many cases, we used it for user-facing io but not much else.  It was stagnant for seemingly quite a while, too.

Deledrius: I believe compilation can be achieved on Linux by adding `#include <memory>` in the files using `std::make_unique`.

Jrius: > Not sure about your experience, but when I was in school it was mostly true, to my recollection.

Gotta admit, C++11 wasn't widely used or supported back when I first learnt it using online tutorials. A few years later I had proper courses at school, but most students never actually programmed before that, so teachers spent most of their time explaining what pointers are. That was probably around the release of C++14, and I didn't have much use for C++ after that, so yeah.

Deledrius: C++11 was in the distant future when I was in school. ;)

core/PRP/Geometry/plDrawableSpans.h
```diff
@@ -27,6 +27,7 @@
 #include "plGeometrySpan.h"
 #include <vector>
 #include <list>
+#include <memory>
```
Hoikas, line 30: This include should be moved to the .cpp file

core/PRP/Region/hsBounds.cpp
```diff
@@ -176,6 +176,16 @@ hsBounds3Corners hsBounds3::getCorners() const
     return corners;
 }
 
+void hsBounds3::setFromPoints(size_t numPoints, const hsVector3* points)
+{
+    if (numPoints > 0) {
+        setMins(points[0]);
+        setMaxs(points[0]);
+    }
```
Hoikas, line 184: It seems to me that calling this method with `numPoints <= 0` is not useful, so I think the if statement could be removed.

Jrius: Wasn't sure whether this might be acceptable when dealing with exotic Blender scene setup. But yeah, probably better to let the caller handle zero-vertices cases.

Hoikas (approved):

zrax (approved):

PR #209, opened 2021-05-30
Eax listener bindings

Jrius:
Python bindings for EAX related stuff.

There is still a compilation error on Linux/macOS, which I would like help with. In `pyEaxListenerMod.cpp`, I need to expose an `unsigned long` attribute to Python. This required adding in `PyPlasma.h` a new overload of the `pyPlasma_xxx` methods, but this clashes with the `size_t` version of same methods for those compilers. I'm not sure how best to solve it.

Also, `EAXREVERBPROPERTIES` is located in an external class in `3rdPartyLibs/AL/EFX-Util.h`. So I just made it a value type inside `pyEAXListenerMod.h/.cpp`. Feel free to suggest improvements.

Example usage:
```
import PyHSPlasma as pl
rmgr = pl.plResManager()
age = rmgr.ReadAge(r"C:\Path\To\Uru\dat\Ahnonay.age", True)
# EAX reverb regions
keys = rmgr.getKeys(rmgr.getLocations()[2], pl.plFactory.kEAXListenerMod)
listener = keys[0].object
print(listener.softRegion)
print(listener.listenerProps.hfReference)
print(listener.listenerProps.airAbsorptionHF)
listener.listenerProps = pl.EAXReverbProperties.REVERB_PRESET_SEWERPIPE
print(listener.listenerProps.hfReference)
print(listener.listenerProps.airAbsorptionHF)
# EAX-enabled sounds
keys = rmgr.getKeys(rmgr.getLocations()[2], pl.plFactory.kWin32StreamingSound)
soundemit = keys[3].object
print(soundemit.eaxSettings.enabled)
print(soundemit.eaxSettings.airAbsorptionFactor)
print(soundemit.eaxSettings.softStarts.directRatio)
```


Python/CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ set(PRP_AUDIO_SOURCES
     PRP/Audio/py2WayWinAudible.cpp
     PRP/Audio/pyAudible.cpp
     PRP/Audio/pyAudibleNull.cpp
+    PRP/Audio/pyEAXEffects.cpp
+    PRP/Audio/pyEAXListenerMod.cpp
```
Hoikas, line 107: The file is misnamed as `pyEaxListenerMod.cpp`, which will error on case sensitive file systems, such as Linux and macOS.

Python/CMakeLists.txt
```diff
@@ -114,6 +116,8 @@ set(PRP_AUDIO_SOURCES
 )
 set(PRP_AUDIO_HEADERS
     PRP/Audio/pyAudible.h
+    PRP/Audio/pyEAXEffects.h
+    PRP/Audio/pyEAXListenerMod.h
```
Hoikas, line 120: The file is misnamed as `pyEaxListenerMod.h`, which will error on case sensitive file systems, such as Linux and macOS.

Hoikas (commented):
Please add [PEP 484 type hints](https://www.python.org/dev/peps/pep-0484/) to `Python/PyHSPlasma.pyi`

Python/CMakeLists.txt
Jrius: Yep, Git didn't catch it because windoz. Thanks for the heads up.

Deledrius:
This looks great, aside from the already-mentioned corrections. 👍🏼

Jrius:
Okay, the stubtest tool appears to be satisfied with my modifs to the PYI file. It doesn't look like it checks it against PyHSPlasma.pyd though. Is there a better way to ensure the file is valid ?

Hoikas:
I'm going to defer to the type master (aka @Zrax) on the `unsigned long` business, but I think the best way to fix that would be to change the storage type to `uint32_t`.

Python/PyPlasma.h
```diff
@@ -356,10 +360,12 @@ template <> inline unsigned char pyPlasma_get(PyObject* value) { return (unsigne
 template <> inline uint16_t pyPlasma_get(PyObject* value) { return (uint16_t)(unsigned long)PyInt_AsLong(value); }
 template <> inline uint32_t pyPlasma_get(PyObject* value) { return (uint32_t)(unsigned long)PyInt_AsLong(value); }
 template <> inline uint64_t pyPlasma_get(PyObject* value) { return (uint64_t)(unsigned long)PyInt_AsLong(value); }
+template <> inline unsigned long pyPlasma_get(PyObject* value) { return (unsigned long)PyInt_AsLong(value); }
 template <> inline signed char pyPlasma_get(PyObject* value) { return (signed char)PyInt_AsLong(value); }
 template <> inline int16_t pyPlasma_get(PyObject* value) { return (int16_t)PyInt_AsLong(value); }
 template <> inline int32_t pyPlasma_get(PyObject* value) { return (int32_t)PyInt_AsLong(value); }
 template <> inline int64_t pyPlasma_get(PyObject* value) { return (int64_t)PyInt_AsLong(value); }
+template <> inline long pyPlasma_get(PyObject* value) { return (long)PyInt_AsLong(value); }
```
zrax, line 368: You can't really mix `int**_t` and `long` safely in overloads, as you've no doubt discovered...  Probably the safest fix (and the one I ended up going with for `string_theory`) is to use the built-in types exclusively.  That is, update all the overloads from `[u]int##_t` to `[unsigned] short/int/long/long long`.  You'll also want to be sure to update the casts where appropriate.

core/3rdPartyLibs/AL/EFX-Util.h
```diff
@@ -420,3 +420,22 @@ void ConvertOcclusionParameters(EAXOCCLUSIONPROPERTIES *pOcProp, EFXLOWPASSFILTE
...
+// Reverberation level
+constexpr unsigned long REVERB_FLAG_REVERB_SCALE = 0x8;
+// Late reverberation delay time
+constexpr unsigned long REVERB_FLAG_REVERB_DELAY_SCALE = 0x10;
+// Echo time
+constexpr unsigned long REVERB_FLAG_ECHO_TIME_SCALE = 0x40;
+// Modulation time
+constexpr unsigned long REVERB_FLAG_MODULATION_TIME_SCALE = 0x80;
+// Limit unnaturally long decay times of high-frequency sounds by forcing a limit to the decay time to be calculated from the Air Absorption HF value
+constexpr unsigned long REVERB_FLAG_DECAY_HF_LIMIT = 0x20;
```
zrax, line 441: Where did these constants come from?

This header file comes from OpenAL originally (note the "3rd party" in the path), and I'm fairly certain the original is a C-compatible header, meaning it wouldn't use `constexpr`.  I'd prefer not to customize a third party header, so this should either be updated to match a version of the OpenAL header, or be put into another appropriate header file.

Python/PyPlasma.h
Jrius: Hopefully that won't cause issues elsewhere in the bindings. Okay, I'll try that soon.

core/3rdPartyLibs/AL/EFX-Util.h
Jrius: Ah, right. My bad, that's a hacky copy-paste from another OpenAL file. I forgot to fix this before the PR, will do in the next commit.

Jrius: Okay, after more digging, those flags aren't defined in OpenAL at all. They come from a file named `eax.h` which *appears* to be from the original DirectSound3D library - but I can't find where to download it anymore. The `eax.h` file is redistributed in various GitHub projects, though. Most notably [doom3-gpl](https://dev.visucore.com/doom3/doxygen/eax_8h_source.html) (where it's GPL-v3 licensed) and [GTA RE3](https://git.pocketdevs.org/Substitute/Re3/src/branch/master/src/audio/eax/eax.h) (unlicensed).

We could add `eax.h` as a new third-party file, although I'm not really fond of the idea because of the redundant types it declares and the need to fix the includes. Or we could just keep the flags we're interested in and put them in another header, whether an existing one like `plEAXListenerMod.h`, or a new header somewhere (in that case, do you have a preference for the location/naming of the file ?).

I'll probably squash/force push some of the commits afterwards in order to remove my fiddlings from the history of EFX-Util.h.

Jrius:
Hm, more errors. Bah, I should just take the time to setup HSPlasma on my Linux machine, will be easier to debug...

Python/PRP/Audio/pyEAXListenerMod.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+    else if (pyEAXReverbProperties_Check(value)) {
+        eaxrp = *((pyEAXReverbProperties*)value)->fThis;
+        self->fThis->setListenerProps(eaxrp);
+        return 0;
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "listenerProps must be a EAXReverbProperties");
+        return -1;
+    }
```
zrax, line 51:
```suggestion
    if (value == Py_None) {
         self->fThis->setListenerProps({});
         return 0;
    } else if (pyEAXReverbProperties_Check(value)) {
        self->fThis->setListenerProps(*((pyEAXReverbProperties*)value)->fThis);
        return 0;
    } else {
        PyErr_SetString(PyExc_TypeError, "listenerProps must be a EAXReverbProperties");
        return -1;
    }
```

Python/PRP/Audio/pyEAXListenerMod.cpp
```diff
@@ -0,0 +1,278 @@
...
+    pyEAXReverbProperties_roomRolloffFactor_getset,
+    pyEAXReverbProperties_flags_getset,
+    PY_GETSET_TERMINATOR
+};
+
+PY_PLASMA_TYPE(EAXReverbProperties, EAXREVERBPROPERTIES, "EAXREVERBPROPERTIES wrapper")
+
+#define PY_TYPE_ADD_PRESET(preset) \
+    PyDict_SetItemString(pyEAXReverbProperties_Type.tp_dict, #preset, \
+        pyEAXReverbProperties_FromEAXReverbProperties(EAXREVERBPROPERTIES(##preset##)));
```
zrax, line 138:
```suggestion
        pyEAXReverbProperties_FromEAXReverbProperties(EAXREVERBPROPERTIES(preset)));
```

zrax:
This will now need to be rebased due to #213, which should fix up the `pyPlasma_convert` stuff

Jrius:
> 
> 
> This will now need to be rebased due to #213, which should fix up the `pyPlasma_convert` stuff

Great ! Will do this soon. I'll use the opportunity to squash all my commits, to remove the unnecessary fluff.

PR #212, opened 2021-06-07
Improve PRP read performance.

Hoikas:
I tested this by loading the Garrison.age from MOULa. In the beginning, loading Garrison took approximately 34 seconds. After these changes, the time is down to approximately 6.5 seconds to load the Age. The major improvement comes from bypassing many calls to `ftell()`, which tank the performance on Windows. The downside is that this makes `pfSizedStream` a bit less flexible in that it assumes that it owns the underlying base stream and no one else will touch it while it lives.

Benchmark tool I used can be found [here](https://github.com/Hoikas/libhsplasma_bench).

### DIFF
--- a/core/Stream/pfSizedStream.cpp
+++ b/core/Stream/pfSizedStream.cpp
@@ -18,7 +18,7 @@
 #include "Debug/plDebug.h"
 
 pfSizedStream::pfSizedStream(hsStream* S, uint32_t len)
-    : fBase(S), fLength(len)
+    : fBase(S), fLength(len), fPos()
 {
     if (S) {
         ver = S->getVer();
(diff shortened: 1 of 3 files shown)


core/Stream/pfSizedStream.cpp
```diff
@@ -41,37 +41,43 @@ void pfSizedStream::seek(uint32_t pos)
     }
 
     fBase->seek(fBegin + pos);
+    fPos = pos > fLength ? fLength : pos;
```
zrax, line 44: The check above is just a warning, so I feel like we should just set `fPos = pos` directly here, even though that puts it out of range.  More importantly, doing so more closely matches the behavior of the code before these changes.

zrax (approved):

PR #213, opened 2021-06-07
Clean up PyHSPlasma integer wrappers

zrax:
* Use built-in types for conversion rather than [u]int{8,16,32,64}_t
* Remove no-longer-necessary `size_t` overload
* For Python 2.x, automatically select `int` or `long` type when required (`int` is still preferred where possible) when converting values from C++
* For Python 2.x, also accept `long` values wherever `int` is accepted.


Hoikas (approved):
Part of me wonders how much we really need to retain Python 2 support since Python 2.7's sunset, but otherwise :+1:

zrax:
I considered just dropping Python 2.x support, but I admit I don't know who still needs it...  I guess one way to find out would be to drop it ;).

PR #214, opened 2021-06-07
Drop support for Python 2.x and 3.0-3.1

zrax:
Drop support for Python versions prior to 3.2, which simplifies the bindings a bit.

This also ports the included scripts to Python 3.x syntax, although the script changes are not tested.

### DIFF
--- a/Python/PyPlasma.h
+++ b/Python/PyPlasma.h
@@ -53,7 +53,7 @@ struct pySequenceFastRef
     PyObject* get(Py_ssize_t idx) { return PySequence_Fast_GET_ITEM(fObj, idx); }
 };
 
-#if (PY_MAJOR_VERSION < 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 3))
+#if (PY_MAJOR_VERSION < 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 2))
     #error Your Python version is too old.  Only 3.2 and later are supported
 #endif
 
--- a/Python/PyPlasma.h
+++ b/Python/PyPlasma.h
@@ -23,7 +23,6 @@
 #include <string_theory/string>
 #include <type_traits>
 
-PyObject* PyString_FromSTString(const ST::string& str);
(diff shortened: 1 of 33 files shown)


Python/scripts/wxSoundDecompress.py
```diff
@@ -437,9 +437,12 @@ def run(self):
...
                                     channelOptions = {}
 
-                                if (soundBuffer.flags & soundBuffer.kOnlyRightChannel): channel = channelOptions["Right"] = True
-                                if (soundBuffer.flags & soundBuffer.kOnlyLeftChannel): channel = channelOptions["Left"] = True
-                                if channelOptions == {}: channelOptions["Both"] = True
+                                if (soundBuffer.flags & soundBuffer.kOnlyRightChannel):
+                                    channel = channelOptions["Right"] = True
+                                if (soundBuffer.flags & soundBuffer.kOnlyLeftChannel):
+                                    channel = channelOptions["Left"] = True
+                                if channelOptions == {}:
```
Hoikas, line 444:
```suggestion
                                if not channelOptions:
```

Hoikas (approved):

Python/PyPlasma.h
```diff
@@ -54,42 +53,8 @@ struct pySequenceFastRef
...
-    #define PyBytes_AsString PyString_AsString
-    #define PyBytes_AS_STRING PyString_AS_STRING
-#else
-    #error Your Python version is too old.  Only 2.6 and later are supported
-#endif
-
-// The type of hashfunc's return value changed in Python 3.2
-#if (PY_MAJOR_VERSION < 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 2))
-    typedef long Py_hash_t;
+#if (PY_MAJOR_VERSION < 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 3))
```
dgelessus, line 56: The condition doesn't match the message - it should probably be
```suggestion
#if (PY_MAJOR_VERSION < 3) || ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 2))
```

PR #215, opened 2021-06-10
Add support for waveset buoys

dpogue:
This depends on https://github.com/H-uru/Plasma/pull/928 to add the equivalent support for runtime buoys to the engine itself.


Hoikas (commented):
Please add [PEP 484](https://www.python.org/dev/peps/pep-0484/) type hints to PyHSPlasma.pyi.

Python/PRP/Surface/pyWaveSet7.cpp
```diff
@@ -94,13 +94,52 @@ PY_METHOD_VA(WaveSet7, delDecal,
...
+    self->fThis->clearBuoys();
+    Py_RETURN_NONE;
+}
+
+PY_METHOD_VA(WaveSet7, delBuoy,
+    "Params: idx\n"
+    "Remove a buoy")
+{
+    int idx;
+    if (!PyArg_ParseTuple(args, "i", &idx)) {
```
Hoikas, line 121:
```suggestion
    Py_ssize_t idx;
    if (!PyArg_ParseTuple(args, "n", &idx)) {
```

Hoikas (approved):

PR #216, opened 2021-06-12
EAXListenerMod bindings (improved)

Jrius:
Rebased #209 on 3460b08f782c9faa7b2221e1e93d5f9e3d8e9763 (thanks for the help Zrax !)
Implemented Zrax's suggestions, moved the problematic flags to a new header file. Squashed all commits to clean up history of my fumblings.

Co-authored-by: Michael Hansen <zrax0111@gmail.com>
Co-authored-by: Adam Johnson <AdamJohnso@gmail.com>


Python/PRP/Audio/pyEAXEffects.cpp
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with HSPlasma.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "pyEAXEffects.h"
+
+#include <PRP/Audio/plEAXEffects.h>
+#include "PRP/Modifier/pyModifier.h"
+#include "PRP/KeyedObject/pyKey.h"
```
Hoikas, line 21:
```suggestion
```

I don't see any uses of these headers in the code?

Python/PRP/Audio/pyEAXEffects.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#ifndef _PYEAXEFFECTS_H
+#define _PYEAXEFFECTS_H
+
+#include "PyPlasma.h"
+#include <PRP/Audio/plEAXEffects.h>
+
+PY_WRAP_PLASMA(EAXSourceSettings, plEAXSourceSettings);
+PY_WRAP_PLASMA(EAXSourceSoftSettings, plEAXSourceSoftSettings);
```
Hoikas, line 24: I wonder if these should use value wrappers since they are not stored as pointers on the sound objects?

Python/PRP/Audio/pyEAXEffects.cpp
Jrius: Yep, good catch.

Python/PRP/Audio/pyEAXEffects.h
Jrius: Maybe. I used the wrapper code for `plSound::plFadeParams` (`pySound.h`) as template for those classes. But wrapping those as value types would be more consistent, I guess ?

Hoikas: *Realizes he never answered here*... Maybe, but also it's good for things to behave consistently, so I guess I'm inclined to leave it alone for now.

Hoikas:
Is there anything left that needs to be done here? A little selfish, but I just realized I needed to use PyHSPlasma to edit some `plEAXSourceSettings`s 😄

zrax (approved):

PR #229, opened 2021-10-19
Add missing server messages

roblabla:
libhsplasmanet is missing three messages from H-uru/Plasma: kCli2Auth_ScoreGetHighScores, kAuth2Cli_ScoreGetHighScoresReply and kAuth2Cli_ServerCaps. The later one in particular prevents it from connecting to dirtsand servers, breaking tools like MoulKI. This MR implements all three, which fixes the issue.


dpogue:
Ironically, `kAuth2Cli_ServerCaps` is a H'uru invention that is only supported by Dirtsand 😛

zrax (approved):

PR #231, opened 2021-11-05
Add missing GeoSpan members to PRC

### LINKED ISSUE #230: Shared Mesh issue
Shared mesh for clothing when modified with PRPShop will disappear if you get too close to the clothing in game.
I found when doing a binary compare of before and after that the initial section of the binary was very different.
Replacing the initial section with the original changes would fix the disappearing clothes.

I think that the maxbones and penbones are not being saved correctly, or the structure of the file is somehow important to MOULA in displaying the clothing correctly. 
This only affects the visibility of the clothing when you get close to them.

image

dpogue:
Fixes #230.


zrax (approved):

Deledrius (approved):
Looks good to me!

PR #232, opened 2021-11-16
Use CMake's GenerateExportHeader module instead of rolling our own.

zrax:
This also enables proper symbol visibility control in GCC, where it was previously only used in MSVC.


CMakeLists.txt
```diff
@@ -4,8 +4,10 @@ project(libhsplasma)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```
Hoikas, line 8: I feel like these should be set per-target so as to not conflict with projects that may want to include HSPlasma as a sub-project (eg via `FetchContent_MakeAvailable()`). But, seeing as how we already clobber the `CMAKE_CXX_STANDARD`, I don't think this is a deal breaker.

core/CMakeLists.txt
```diff
@@ -817,6 +816,9 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.16)
     set_property(SOURCE ${SQUISH_SOURCES} PROPERTY SKIP_UNITY_BUILD_INCLUSION TRUE)
 endif()
 
+include(GenerateExportHeader)
+generate_export_header(HSPlasma)
```
Hoikas, line 820: Did you consider using `EXPORT_MACRO_NAME PLASMA_DLL` to reduce the amount of noise in the diff?

zrax: I did strongly consider it for exactly that reason...  But a) I like `HSPLASMA_EXPORT` better, especially considering it now applies to non-DLL shared objects as well, and b) it's more consistent with the usual pattern for using this CMake module.

PR #233, opened 2021-11-30
Fix incorrect keyring flag.

Hoikas:
Legacy MOUL clients expect for the keyring to be sorted by name. If keys are not sorted and the flag is not applied, then searches by name will silently fail. This was observed when someone attempted to add a new image to GUI_District_BkBookImages.prp and use it a journal. Due to this issue, no image was returned by the journal's search.


zrax:
If searching is more efficient with a sorted keyring, perhaps we should sort the keyring instead...

Hoikas:
Perhaps, but that codepath was actually removed from H'uru/Plasma, so we'd be writing more code to support a dead-end (MOULa). I'm also wary of introducing another case similar to #218.

zrax:
True, but we also support several other dead-end games, and this code is in a block that would also affect Myst 5, for example...

Hoikas:
That's a good point. To keep the code simple, I wonder if it would make sense to make the optimized sort an explicit step that needs to be done before writing out a page? Or would you prefer to maintain the key list in its optimized form at all times? My concern with doing that is that there will be more potential for keys to be re-numbered when doing seemingly trivial operations.

dpogue:
> My concern with doing that is that there will be more potential for keys to be re-numbered when doing seemingly trivial operations.

After #218 aren't the keys always re-numbered when doing any writing at all? If that's the case, it doesn't seem like it should be too difficult to sort them while re-numbering them

Hoikas:
#218 changed it to re-number the keys on read. Previously, we were re-numbering them on write, which would silently break the world. Note that's re-number, as in, make the values contiguous, starting at 1. For this keyring flag to be zero, we need to case-insensitve sort the keys, which is an even more fun operation ;)

Hoikas:
Ok, I've added sort-by-name as an opt-in feature with mitigation for old (read: broken) files produced by previous revisions of libHSPlasma.

Python/ResManager/pyResManager.cpp
```diff
@@ -476,6 +476,21 @@ PY_METHOD_VA(ResManager, getKeys,
...
 
+PY_METHOD_VA(ResManager, optimizeKeys,
+    "Params: location\n"
+    "Sorts keys in the keyring by name. This is probably a good idea\n"
+    "to do if you're changing a PRP for MOUL or Myst V that has objects\n"
+    "(such as journal images) that are looked up by name.")
+{
+    pyLocation* loc;
+    if (!PyArg_ParseTuple(args, "O", &loc) || !!pyLocation_Check((PyObject*)loc)) {
+        PyErr_SetString(PyExc_TypeError, "optimizeKeys expects plLocation");
```
zrax, line 487:
```suggestion
        PyErr_SetString(PyExc_TypeError, "optimizeKeys expects a plLocation");
```

core/ResManager/plKeyCollector.cpp
```diff
@@ -141,21 +142,58 @@ void plKeyCollector::cleanupKeys()
...
     for (short type : types) {
-        std::sort(keys[loc][type].begin(), keys[loc][type].end(),
+        std::sort(keys[loc][type]->begin(), keys[loc][type]->end(),
             [](const plKey& a, const plKey& b) { return a->getID() < b->getID(); });
-        for (size_t i = 0; i < keys[loc][type].size(); ++i) {
-            keys[loc][type][i]->setID(i + 1);
+        for (size_t i = 0; i < keys[loc][type]->size(); ++i) {
+            size_t newID = i + 1;
+            if (keys[loc][type][i]->getID() != newID)
+                keys[loc][type][i]->setID(newID);
```
zrax, line 167: Is there any need to check?  It looks like this just does a simple assignment...
```suggestion
            keys[loc][type][i]->setID(newID);
```

core/ResManager/plKeyCollector.cpp
```diff
@@ -141,21 +142,58 @@ void plKeyCollector::cleanupKeys()
...
+
+            // Side effect: verify that pages are telling the truth about
+            // whether or not they are optimized. Fixes incorrect pages
+            // exported by previous revisions of HSPlasma.
+            if (!keys[loc][type].checkFlag(kNotOptimized) && newID < keys[loc][type]->size()) {
+                const plKey& a = keys[loc][type][i];
+                const plKey& b = keys[loc][type][newID];
```
zrax, line 174: Because of these lines, I may have a slight preference for using `i + 1` directly instead of storing it in `newID`...  Or maybe call it `nextID` so it's more clear that it comes from the same sequence as `i` rather than some other random index.

core/ResManager/plKeyCollector.h
```diff
@@ -21,7 +21,42 @@
...
+            fFlags &= ~flag;
+    }
+
+    void setFlags(uint32_t flags) { fFlags = flags; }
+
+    std::vector<plKey>* operator ->() { return &fKeys; }
+    const std::vector<plKey>* operator ->() const { return &fKeys; }
+
+    std::vector<plKey>& operator *() { return fKeys; }
+    const std::vector<plKey>& operator *() const { return fKeys; }
```
zrax, line 53: I might prefer to wrap the vector methods we need, rather than make this look like a pointer type (which it isn't)...

core/ResManager/plKeyCollector.cpp
```diff
@@ -183,7 +219,7 @@ std::vector<plKey> plKeyCollector::getKeys(const plLocation& loc, short type,
         }
         return std::vector<plKey>(kList.begin(), kList.end());
     } else {
-        return keys[loc][type];
+        return *keys[loc][type];
```
zrax, line 222:
```suggestion
        return keys[loc][type].toStdVector();
```

core/ResManager/plKeyCollector.h
```diff
@@ -21,7 +21,48 @@
...
+    bool empty() const HS_NOEXCEPT { return fKeys.empty(); }
+    std::vector<plKey>::iterator end() HS_NOEXCEPT { return fKeys.end(); }
+    std::vector<plKey>::iterator erase(std::vector<plKey>::iterator pos) { return fKeys.erase(pos); }
+    void push_back(const plKey& value) { fKeys.push_back(value); }
+    void reserve(size_t count) { fKeys.reserve(count); }
+    void resize(size_t count) { fKeys.resize(count); }
+    size_t size() const HS_NOEXCEPT { return fKeys.size(); }
+
+    std::vector<plKey>& operator *() { return fKeys; }
+    const std::vector<plKey>& operator *() const { return fKeys; }
```
zrax, line 59:
```suggestion
    std::vector<plKey>& toStdVector() { return fKeys; }
    const std::vector<plKey>& toStdVector() const { return fKeys; }
```

zrax (approved):

PR #235, opened 2022-01-01
Remove windows-2016 runner from CI.

Hoikas:
The windows-2016 runner will be removed in just a few short months. To retain CI coverage with VS 2017, I've changed those jobs to use VS 2019 but with the v141 toolset. Additionally, I've added a new job for VS 2022 and Python 3.10.


Hoikas:
We have just under a month before windows-2016 is removed.

PR #239, opened 2022-02-05
Fix invalid format string.

Hoikas:
Whoops.

### DIFF
--- a/core/ResManager/plKeyCollector.cpp
+++ b/core/ResManager/plKeyCollector.cpp
@@ -171,7 +171,7 @@ void plKeyCollector::sortKeys(const plLocation& loc)
                 const plKey& a = keys[loc][type][i];
                 const plKey& b = keys[loc][type][i + 1];
                 if (!(a->getName().compare_i(b->getName()) < 0)) {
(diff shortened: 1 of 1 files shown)


core/ResManager/plKeyCollector.cpp
```diff
@@ -171,7 +171,7 @@ void plKeyCollector::sortKeys(const plLocation& loc)
                 const plKey& a = keys[loc][type][i];
                 const plKey& b = keys[loc][type][i + 1];
                 if (!(a->getName().compare_i(b->getName()) < 0)) {
-                    plDebug::Warning("Keyring for page {} type {:04X} erroneously thinks it's optimized ({} >= {})",
+                    plDebug::Warning("Keyring for page {} type {04X} erroneously thinks it's optimized ({} >= {})",
```
zrax, line 174: In retrospect, not making ST format strings closer to Python's was probably a mistake...  But one that we can't really undo at this point without breaking everything :/

zrax (approved):

core/ResManager/plKeyCollector.cpp
Hoikas: Yeah... I wrote this completely on auto-pilot. Maybe ST should ignore any colons in the format string so that Python style format strings would work?

PR #241, opened 2022-02-19
Don't touch my `CMAKE_` settings!

Hoikas:
Many `CMAKE_` variables are considered to be user-facing settings. It's OK to set default values for these inside the project, but we have been using them to set values required to compile HSPlasma. This is problematic if someone wants to, for example, use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) to include HSPlasma as a first class citizen of their project (ie as a subproject). Specifically the old `set(CMAKE_CXX_STANDARD 14)` line would blow away any setting of the default C++ standard. This is not a good UX, so this pull requests removes all `CMAKE_` forced settings into target properties as expected in "modern" CMake.

As bonuses, this corrects some unnecessary virality of our 3rd party dependencies and fixes an issue with string_theory not being pulled in by consumers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
 project(libhsplasma)
 
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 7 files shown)


Hoikas:
I'm going to break this PR down into topics to facilitate review. This one is stale, anyhow.

PR #242, opened 2022-03-04
Use Win32 API in `hsFileStream`.

Hoikas:
When developing Korman's "Test Age" feature, I noticed a curious bug that I could never quite understand. The feature basically uses `hsFileStream` to modify the TPotS vault.dat, then launches Uru. Occassionally, however, Uru would act as if the changes to the vault were not made - but according to VaultShop, the changes were present! I discovered that adding [a small sleep](https://github.com/H-uru/korman/blob/1e198a1cfe725bceaa090ee3f910608cce97f1e9/korman/plasma_launcher.py#L177:L178) between writing vault.dat and launching Uru made the process slightly more reliable, but still not 100%.

As it turns out, the C file API in Windows is evidentally vulnerable to race conditions where changes are not flushed out as expected. I'm not mthe only one to discover this problem: https://twitter.com/jonathan_blow/status/1417544504916135936

### DIFF
--- a/core/Stream/hsStream.cpp
+++ b/core/Stream/hsStream.cpp
@@ -21,7 +21,7 @@
 #include "Sys/Platform.h"
 #include "Debug/plDebug.h"
 
-#ifdef WIN32
+#ifdef _WIN32
     #define WIN32_LEAN_AND_MEAN
     #define NOMINMAX
     #include <windows.h>
@@ -353,7 +353,7 @@ void hsStream::writeLine(const ST::string& ln, bool winEOL)
 /* hsFileStream */
 bool hsFileStream::FileExists(const ST::string& file)
 {
-#ifdef WIN32
+#ifdef _WIN32
     HANDLE eFile = CreateFileW(file.to_wchar().data(), GENERIC_READ,
         FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr);
     if (eFile == INVALID_HANDLE_VALUE)
--- a/core/Stream/hsStream.cpp
+++ b/core/Stream/hsStream.cpp
@@ -371,7 +371,7 @@ bool hsFileStream::FileExists(const ST::string& file)
 
 bool hsFileStream::open(const ST::string& file, FileMode mode)
 {
-#ifdef WIN32
+#ifdef _WIN32
     DWORD access, disposition;
     switch (mode) {
     case FileMode::fmRead:
@@ -383,7 +383,7 @@ bool hsFileStream::open(const ST::string& file, FileMode mode)
         disposition = OPEN_ALWAYS;
         break;
     case FileMode::fmCreate:
-        access = GENERIC_WRITE;
+        access = GENERIC_READ | GENERIC_WRITE;
         disposition = CREATE_ALWAYS;
         break;
     case FileMode::fmReadWrite:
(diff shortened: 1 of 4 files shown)


core/Stream/hsStream.cpp
```diff
@@ -381,83 +430,128 @@ bool hsFileStream::open(const ST::string& file, FileMode mode)
...
+#endif
 }
 
 bool hsFileStream::eof() const
 {
     if (F == nullptr)
         return true;
+
+#ifdef WIN32
+    return pos() == size();
```
zrax, line 482: Hmm, the Posix implementation worked even for files whose size was unknown...  Arguably, that may never be relevant to libhsplasma, but I wonder if there's a way to preserve those semantics here.

core/Stream/hsStream.h
```diff
@@ -87,7 +87,11 @@ class HSPLASMA_EXPORT hsStream
 
 class HSPLASMA_EXPORT hsFileStream : public hsStream {
 protected:
+#ifdef WIN32
+    HANDLE F;
```
zrax, line 91: I don't love having a `HANDLE` in a header file, especially since we've managed to avoid it everywhere else...  Perhaps we could just store a generic `void* F` and then cast it as necessary within `hsStream.cpp`?

zrax (commented):
Thanks, I hate it :(.  But that's MS's fault, not yours...

core/Stream/hsStream.cpp
```diff
@@ -347,15 +353,58 @@ void hsStream::writeLine(const ST::string& ln, bool winEOL)
 /* hsFileStream */
 bool hsFileStream::FileExists(const ST::string& file)
 {
+#ifdef WIN32
+    HANDLE eFile = CreateFileW(file.to_wchar().data(), GENERIC_READ,
+        FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr);
```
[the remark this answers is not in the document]
Hoikas: I was thinking here that it wasn't really important to lock the file for writing as much as it would be in `hsFileStream::open()`. I wouldn't mind changing either this or `hsFileStream::open()` though.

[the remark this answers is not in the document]
zrax: Makes sense 👍

core/Stream/hsStream.h
Hoikas: It looks like MSVC is happy to accept a `void*` without casting any time a function wants a `HANDLE`. That's nice.

core/Stream/hsStream.h
```diff
@@ -87,7 +87,11 @@ class HSPLASMA_EXPORT hsStream
 
 class HSPLASMA_EXPORT hsFileStream : public hsStream {
 protected:
+#ifdef WIN32
+    void* F;
+#else
```
zrax, line 92:
```suggestion
#ifdef _WIN32
    /* HANDLE */ void* F;
#else
```
Just so it's clear what we're actually doing here...  >.>

core/Stream/hsStream.cpp
```diff
@@ -21,6 +21,12 @@
 #include "Sys/Platform.h"
 #include "Debug/plDebug.h"
 
+#ifdef WIN32
```
zrax, line 24:
```suggestion
#ifdef _WIN32
```

core/Stream/hsStream.cpp
```diff
@@ -347,15 +353,59 @@ void hsStream::writeLine(const ST::string& ln, bool winEOL)
 /* hsFileStream */
 bool hsFileStream::FileExists(const ST::string& file)
 {
+#ifdef WIN32
```
zrax, line 356:
```suggestion
#ifdef _WIN32
```

core/Stream/hsStream.cpp
```diff
@@ -381,83 +431,134 @@ bool hsFileStream::open(const ST::string& file, FileMode mode)
...
     if (F == nullptr)
         return true;
+
+#ifdef _WIN32
+    uint8_t buf;
+    DWORD nread = 0;
+    BOOL result = ReadFile(F, &buf, 1, &nread, nullptr);
+    if (result == TRUE && nread == 0)
+        return true;
+    SetFilePointer(F, -1, nullptr, FILE_CURRENT);
```
zrax, line 488: Hmm, this still probably won't work on non-seekable streams...  But I guess the Win32 API doesn't have any way to push a byte back into the read buffer like C stdio does :/.

Oh well, probably good enough for our purposes.

Hoikas: Yeah, the only thing I saw for peeking data from a `HANDLE` was specifically for pipes 😞

zrax (approved):

Issue #23, opened 2012-09-01
getVertices() of plGBufferGroup


Jrius:
Files: core/PRP/Geometry/plGBufferGroup.h and plGBufferGroup.cpp

Issue originally posted at the Guild of Writers: http://forum.guildofwriters.org/viewtopic.php?f=114&t=5872

It seems using getVertices() function of plGBufferGroup doesn't work as expected.
Calling this function with start != 0 crashes the program using HSPlasma.
When setting start to 0 and count to any number, the function returns every plGBufferVertex from (start) to (last index - count), instead of from (start) to (start+count).

I had this issue using the Windows 64-bits r848 Python interface.

PR #243, opened 2022-04-23
Fix pyGBufferGroup.getVertices (#23)

Jrius:
Almost ten years later, I found the reason for #23. An argument was missing due to a simple oversight.


Hoikas (approved):
Whoopsie daisy

Issue #244, opened 2022-05-17
Files not saving correctly with Use Win32 API commit


Hazado:
https://github.com/H-uru/libhsplasma/commit/21c5f671a05994d30449eb7b5360f86b6e5ec118

This commit causes objects in the file to not be removed correctly.
If you remove everything in a prp with PRPShop, the file size will stay the same after you save it.

PR #246, opened 2022-05-18
Revert "Use Win32 API in `hsFileStream`."

zrax:
Reverts H-uru/libhsplasma#242

This causes a massive performance regression in reading PRPs:
* A quick test of loading MOULa's Ahnonay.age with a release build of libhsplasma:
  * With #242: 34sec
  * Without #242: 1.5sec

There are also reports of written PRPs containing the wrong or outdated content (see #244).

This can be re-added in the future if the issues can be resolved, but for now I'd like to get a new release of libhsplasma and PlasmaShop that fixes these regressions.


Hoikas (approved):
As I said in IRC:
> Yeah, we probably need to revert the Win32 stream thing. The overhead on calls to `ReadFile()` is nuts. We'd need to implement our own buffering for this to not suck, and I don't feel like doing that right now...

PR #249, opened 2022-07-16
Python: expose plClothingItem class

Jrius:
... So that ZLZ can finally import clothing items.

(@Hoikas, when preparing for the next release of Korman, could the redistributed HSPlasma.pyd be rebuilt to include this ? Thanks !)


dpogue (approved):
Weird that the implementation already exists but was never hooked up here.

zrax (approved):

PR #250, opened 2022-07-16
Fix for PyInit_PyHSPlasma visibility on Py <3.9

### LINKED ISSUE #248: CMake visibility changes break PyHSPlasma on Python < 3.9
Essentially, `PyInit_PyHSPlasma` isn't marked as visible, so the module initialization fails.

This is fixed in Python >= 3.9 because they added visibility attributes to the `PyMODINIT_FUNC` macro, however some of us are still building with Python 3.7 to run korman on Linux with Blender 2.79 (which depends on Python 3.7).

Based on how libHSPlasma and libHSPlasmaNet are doing things, I imagine there's some sort of `PYHSPLASMA_EXPORT` thing that could be added here to make it visible, but I don't know enough about where and how CMake is defining these macros to be sure of that...

dpogue:
libHSPlasma has been configured to use proper symbol visibility, which has resulted in all of the PyHSPlasma symbols being marked hidden.
`PyMODINIT_FUNC` on Python 3.9+ has been updated to include visibility annotations, but on earlier versions it is missing.

I've confirmed this is working for me when building for Python 3.7.

Closes #248

### DIFF
--- a/Python/Module.cpp
+++ b/Python/Module.cpp
@@ -455,6 +455,10 @@ static void initPyHSPlasma_Constants(PyObject* module)
     PyModule_AddIntConstant(module, "KEY_UNMAPPED", KEY_UNMAPPED);
 }
 
+#if !defined(_WIN32) && ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 9))
+PyMODINIT_FUNC PyInit_PyHSPlasma() __attribute__((visibility("default")));
+#endif
+
 PyMODINIT_FUNC PyInit_PyHSPlasma()
 {
     PyObject* module = PyModule_Create(&PyHSPlasma_Module);


Python/Module.cpp
```diff
@@ -455,6 +455,15 @@ static void initPyHSPlasma_Constants(PyObject* module)
     PyModule_AddIntConstant(module, "KEY_UNMAPPED", KEY_UNMAPPED);
 }
 
+#if ((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION < 9)) && defined(__GNUC__)
+#   define BEGIN_EXPORTED_DECLARATION _Pragma("GCC visibility push(default)")
+#   define END_EXPORTED_DECLARATION _Pragma("GCC visibility pop")
```
zrax, line 460: Why not use an attribute (`__attribute__((visibility("default")))`) instead of needing to wrap the whole function body in begin/end macros?  The attribute macro can be placed after `PyMODINIT_FUNC` to ensure it takes priority over whatever `PyMODINIT_FUNC` defines...

dpogue: I tried that and it said the attribute was in the wrong spot and would be ignored.

`PyMODINIT_FUNC` is `extern "C" PyObject*`, and I think it wants the attribute before the `PyObject*` return type

zrax: Ah, yeah, that does make it more annoying...  However, you could also forward-declare it with the attribute at the end:

```c++
PyMODINIT_FUNC PyInit_PyHSPlasma() __attribute__((visibility("default")));

PyMODINIT_FUNC PyInit_PyHSPlasma()
{
    ...
}
```

dpogue: Ahh okay, that's much cleaner.

dpogue: I'm not sure how "safe" it is to reuse `HSPLASMA_EXPORT` outside of the libHSPlasma core module itself, but it seems to work.

zrax: I'm more worried about how "safe" it is to use in a location not documented as supported by CMake, since that macro now comes from CMake's GenerateExportHeader module.

dpogue: This seems like it would be safe to have unconditionally?

```cpp
PyMODINIT_FUNC PyInit_PyHSPlasma() [[gnu::visibility("default")]];
```

Python 3.7 already (as far as I can tell) annotates it with `__declspec(dllexport)` for MSVC

dpogue: Apparently `__attribute__((visibility("default")))` is the only thing that works... `[[gnu::visibility("default")]]` is ignored on non-class types 🙄

zrax (approved):

PR #251, opened 2022-07-16
MQO reading

### LINKED ISSUE #247: MQO prps fail to open correctly
Opening magiquest prps in prpshop will fail to load some of the data correctly.

dpogue:
Based on some work by @Hazado and references from Drizzle, this seems to be able to read all the MQO files. I'm going to say that editing/writing is probably not safe for now.

For most cases, we can rely on the Class Versions map, except for the case of messages included within responders. That necessitated the addition of a `pvMqo` entry to PlasmaVersion and a very specific filename/SeqPrefix check to decide when to use it or not 😞 

This also adds support for the plCloneSpawnModifier class. Currently this is unused in Uru and MOUL, so we'll always try to read an MQO version if it's MOUL-ish data. Given that the class implementation is actually probably broken in Uru, it makes sense for us to update it in H-uru/Plasma to match MQO at some point so that it's actually usable, but that's neither here nor there.

There's a handful of version checks for Choru data that I've been carrying around locally for a few years in here too. I can try to split them out into their own commit if that's a problem.

Closes #247


core/Util/PlasmaVersions.h
```diff
@@ -85,6 +85,7 @@ class HSPLASMA_EXPORT PlasmaVer
         pvPrime     = MAKE_VERSION(2, 0, 63, 11),
         pvPots      = MAKE_VERSION(2, 0, 63, 12),
         pvMoul      = MAKE_VERSION(2, 0, 70,  0),
+        pvMqo       = MAKE_VERSION(2, 0, 70,  1),
```
Hoikas, line 88: H-uru/Plasma is currently 2.0.70.2, FWIW.

dpogue: That ends up not really mattering because that's handled via the ClassVersions stuff.

MQO isn't really 2.0.70.1, we just need something to use as a flag because of the missing ClassVersions for messages :(

core/PRP/Avatar/plMultistageBehMod.cpp
```diff
@@ -28,7 +28,10 @@ void plMultistageBehMod::read(hsStream* S, plResManager* mgr)
 
     fFreezePhys = S->readBool();
     fSmartSeek = S->readBool();
-    fReverseFBControlsOnRelease = S->readBool();
+
+    if (S->getVer() >= MAKE_VERSION(2, 0, 62, 12)) {
+        fReverseFBControlsOnRelease = S->readBool();
+    }
```
zrax: This should probably also be updated in `write()`

core/PRP/KeyedObject/plLoadMask.cpp
```diff
@@ -18,9 +18,13 @@
 
 void plLoadMask::read(hsStream* S)
 {
-    unsigned char m = S->readByte();
-    fQuality[0] = (m >> 4) | 0xF0;
-    fQuality[1] = m | 0xF0;
+    if (S->getVer() < MAKE_VERSION(2, 0, 63, 0) && S->getVer().isValid()) {
+        S->readInt();
```
zrax: Need to update `write()`

core/PRP/Modifier/plOneShotMod.cpp
```diff
@@ -41,6 +45,10 @@ void plOneShotMod::write(hsStream* S, plResManager* mgr)
     S->writeBool(fReversable);
     S->writeBool(fSmartSeek);
     S->writeBool(fNoSeek);
```
zrax: Make this conditional to match `read()`?

core/PRP/Physics/plGenericPhysical.cpp
```diff
@@ -99,6 +99,10 @@ void plGenericPhysical::read(hsStream* S, plResManager* mgr)
 {
     plSynchedObject::read(S, mgr);
 
+    if (S->getVer() < MAKE_VERSION(2, 0, 63, 0)) {
+        return;
+    }
```
zrax: Also update `write()`

dpogue:
I've made those reads/write consistent, and also got support for HKPhysical working.

This now includes a slightly modified version of #138 (in which we still read the hsVector3 because some versions have extra data in there, but assign to the hsQuat without doing any data transformations)

PR #252, opened 2022-07-17
Python: make pySpanTemplateVertex a value type

Jrius:
For some reason `pySpanTemplateVertex` (a Python reference type) sometimes becomes corrupted, which means position, normal etc get filled with junk data. I'm not sure why.
`pyGBufferVertex` which is very similar is a value type. Making `pySpanTemplateVertex` a value type solves that corruption problem on top of being more consistent. :shrug:
(`pySpanTemplateVertex` is used by cluster groups - which ZLZ can now import.)


Hoikas (approved):
This looks correct because this type is used as a value in HSPlasma itself.

zrax (approved):

PR #253, opened 2022-07-23
pyClothingItem index safety and extra properties

Jrius:
`pyClothingItem` was missing a few things to be properly usable, so I made the following modifications:
- Added constants `pyClothingItem.kNumLODLevels` and `pyClothingItem.kLayerMax`.
- Added `pyClothingItem.numElements` getter (as well as matching `plClothingItem::getNumElements()`).
- The following will throw an `IndexError` when appropriate instead of crashing:
    - `getElementTexture`
    - `setElementTexture`
    - `getElementName`
    - `setElementName`
    - `delElement`


core/PRP/Avatar/plClothingItem.h
```diff
@@ -198,6 +198,9 @@ class HSPLASMA_EXPORT plClothingItem : public hsKeyedObject
     /** Add a named element to the clothing item. */
     void addElement(const ST::string& elementName);
 
+    /** Return the number of elements in the clothing item. */
+    int getNumElements() const { return fElementNames.size(); }
```
zrax, line 202: This should return `size_t`

PR #258, opened 2022-12-02
Fix PRC parsing of Particle Source params

### LINKED ISSUE #257: Particle Pitch and Yaw zero out when modified
Making any changes to Particles will cause the Yaw and Pitch to become zero for any particle emitters

Before
image

After
image

dpogue:
This should fix #257.


Hoikas (approved):
CI failure is unrelated to this PR.

zrax (approved):

PR #261, opened 2022-12-03
Only use the SSE2 flag when it's supported by the compiler.

### LINKED ISSUE #260: SSE compile flags unconditionally added
Compiling on macOS for Apple Silicon (Arm64) yields the following warning:

```
clang: warning: argument unused during compilation: '-msse2' [-Wunused-command-line-argument]
```

Also libsquish tries to pull in the SSE headers and fails to compile.

Hoikas:
This should fix #260. [`check_cxx_compiler_flag()`](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) actually tries to compile a dummy file, so the check should fail on ARM targets. This also hides the bogus `ENABLE_SSE2` option when the compiler is MSVC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,26 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
-option(ENABLE_SSE2 "Build with SSE2 CPU instructions" ON)
+include(CheckCXXCompilerFlag)
+include(CMakePushCheckState)
+
+# Note that recent versions of MSVC default to producing SSE2 instructions for x86
+# targets, so it's ok to simply fail this check for MSVC. It will always produce SSE2
+# code for x86 targets in that case.
+set(SSE2_COMPILER_FLAGS "-msse2")
+CMAKE_PUSH_CHECK_STATE()
+set(CMAKE_REQUIRED_FLAGS "-Werror")
+check_cxx_compiler_flag("${SSE2_COMPILER_FLAGS}" SSE2_SUPPORTED)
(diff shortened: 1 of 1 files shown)


dpogue:
Unfortunately, this is detecting SSE2 as supported on Apple Silicon. Maybe because the `-msse2` flag is only a warning in clang so it's not failing the compiler check?

Hoikas:
Hmm... Just to check, are you cross compiling for Apple Silicon on Intel x64? I can do that locally and see what's up.

dpogue:
No, this is compiling on an M1 machine. Terminal, bash, cmake, and clang are all Arm binaries (so there's not Rosetta emulation stuff going on tricking it into thinking it's x86_64)

Hoikas:
I've amended the check to use `-Werror`, which should hopefully solve the warning :)

dpogue (approved):
This seems to work :+1:

CMakeLists.txt
```diff
@@ -9,7 +9,26 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
-option(ENABLE_SSE2 "Build with SSE2 CPU instructions" ON)
+include(CheckCXXCompilerFlag)
+include(CMakePushCheckState)
+
+# Note that recent versions of MSVC default to producing SSE2 instructions for x86
```
zrax, line 15: Just for curiosity's sake, do you happen to know which "recent versions" this applies to?

zrax (approved):

CMakeLists.txt
Hoikas: According to some Googling, it appears that SSE2 became the default in VS 2015.

PR #264, opened 2023-01-16
Add hsWindowsStream.

Hoikas:
While the overhead of the Win32 file IO APIs is absymal without buffering, the original point of #242 remains: sometimes, we need to use the actual Windows APIs to prevent race conditions related to writing to a file, then reading from it in another process. Therefore, we add that functionality, but in a different subclass of `hsStream`.


Hazado:
Tested saving a PRP without any noticeable issues.

Hoikas:
The marking of a draft was in response to an observation when using this in Korman. The problem was in Korman itself, not this PR.

zrax (approved):

PR #266, opened 2023-03-04
Add ResMgr option to preserve Object IDs

dpogue:
After several attempts to address issues around Object ID mismatches, we've adopted a stance of libHSPlasma reassigning all Object IDs on read. This generally works okay (with caveats around cross-page references) but causes problems if you're wanting to round-trip data through libHSPlasma without modifying it.

This adds a `preserveObjectIDs()` method to plResManager that tells it not to change object IDs when reading. This is **not safe** if you are adding or removing any objects from the PRP file, but should be safe if you're only wanting to edit some object data.

This also sets up the `prcdc` tool to always dump out the original Object IDs to get an accurate view of the data.

When running `prcc` to compile from PRC files, there's a new `--keep-ids` option to use the object ID values as read from the PRC. This enables us to round-trip through `prcdc`/`prcc` without renumbering all the data, which should help us to track down any PRC parsing issues.

### DIFF
--- a/core/ResManager/plResManager.h
+++ b/core/ResManager/plResManager.h
@@ -65,6 +65,7 @@ protected:
     LoadProgressCallback progressFunc;
     PageUnloadCallback pageUnloadFunc;
     bool mustStub;
+    bool preserveIDs;
 
 private:
     unsigned int ReadKeyring(hsStream* S, const plLocation& loc);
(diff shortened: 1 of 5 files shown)


Hoikas:
Considering that the reordering is done on read, it would make more sense to me for this to be an off-by-default arguement to `ReadAge()` and `ReadPage()` to skip the call to `sortKeys()`.

core/ResManager/plResManager.h
```diff
@@ -486,6 +487,9 @@ class HSPLASMA_EXPORT plResManager
 
     /** Unlock access to the ResManager in multithreaded apps */
     void unlock() { fResMgrMutex.unlock(); }
+
+    /** Sets a flag to preserve existing Object IDs on read/write. */
+    void preserveObjectIDs() { preserveIDs = true; }
```
zrax, line 492: I think I'd rather see this as an optional parameter in the constructor...  When it's written here, it's a one-way flag anyway (there's no way to unset it), and it would still need to be done before any reads or writes were started.

zrax (approved):

PR #268, opened 2023-03-11
Add new PFM Attribute for plLayers.

Deledrius:
- Adds ptAttribLayer support.

### DIFF
--- a/core/PRP/Modifier/plPythonFileMod.cpp
+++ b/core/PRP/Modifier/plPythonFileMod.cpp
@@ -24,7 +24,7 @@ const char* plPythonParameter::ValueTypeNames[] = {
     "Animation", "AnimationName", "Behavior", "Material", "GUIPopUpMenu",
     "GUISkin", "WaterComponent", "SwimCurrentInterface", "ClusterComponent",
     "MaterialAnimation", "GrassShaderComponent", "GlobalSDLVar", "Subtitle",
-    "BlowerComponent", "None"
+    "BlowerComponent", "Layer", "None"
 };
 
 plPythonParameter& plPythonParameter::operator=(const plPythonParameter& init)
@@ -222,6 +222,15 @@ unsigned int plPythonParameter::PlasmaToMapped(unsigned int type, PlasmaVer ver)
         case 26: return kNone;
         default: return 0;
         }
+    } else if (ver.isMoul()) {
+        switch (type) {
+        case 20: return kClusterComponent;
+        case 21: return kMaterialAnimation;
+        case 22: return kGrassShaderComponent;
+        case 23: return kLayer;
+        case 24: return kNone;
+        default: return 0;
+        }
     } else {
         switch (type) {
         case 20: return kClusterComponent;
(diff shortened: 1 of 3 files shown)


core/PRP/Modifier/plPythonFileMod.h
```diff
@@ -26,10 +26,10 @@ struct HSPLASMA_EXPORT plPythonParameter
     {
         kInt = 1, kFloat, kBoolean, kString, kSceneObject, kSceneObjectList,
         kActivator, kResponder, kDynamicText, kGUIDialog, kExcludeRegion,
-        kAnimation, kAnimationName, kBehavior, kMaterial, kGUIPopUpMenu,
+        kAnimation, kAnimationName, kBehavior, kImage, kGUIPopUpMenu,
```
zrax, line 29: Instead of just renaming this, perhaps we should add an alias to avoid breaking backwards compatibility?

core/PRP/Modifier/plPythonFileMod.cpp
```diff
@@ -250,6 +259,15 @@ unsigned int plPythonParameter::MappedToPlasma(unsigned int type, PlasmaVer ver)
         case kNone: return 26;
         default: return 0;
         }
+    } else if(ver.isMoul()) {
```
zrax, line 262:
```suggestion
    } else if (ver.isMoul()) {
```

core/PRP/Modifier/plPythonFileMod.cpp
```diff
@@ -250,6 +259,15 @@ unsigned int plPythonParameter::MappedToPlasma(unsigned int type, PlasmaVer ver)
...
+        switch (type) {
+        case kClusterComponent: return 20;
+        case kMaterialAnimation: return 21;
+        case kGrassShaderComponent: return 22;
+        case kLayer: return 23;
+        case kNone: return 24;
```
zrax, line 268: Hmm, will this break anything with older MOUL PRPs?

core/PRP/Modifier/plPythonFileMod.cpp
```diff
@@ -222,6 +222,15 @@ unsigned int plPythonParameter::PlasmaToMapped(unsigned int type, PlasmaVer ver)
         case 26: return kNone;
         default: return 0;
         }
+    } else if(ver.isMoul()) {
+        switch (type) {
+        case 20: return kClusterComponent;
+        case 21: return kMaterialAnimation;
+        case 22: return kGrassShaderComponent;
+        case 23: return kLayer;
+        case 24: return kNone;
```
zrax, line 231: Will this break anything with older MOUL PRPs?

core/PRP/Modifier/plPythonFileMod.cpp
```diff
@@ -222,6 +222,15 @@ unsigned int plPythonParameter::PlasmaToMapped(unsigned int type, PlasmaVer ver)
         case 26: return kNone;
         default: return 0;
         }
+    } else if(ver.isMoul()) {
```
zrax, line 225:
```suggestion
    } else if (ver.isMoul()) {
```

core/PRP/Modifier/plPythonFileMod.cpp
Deledrius: Not that I'm aware of, but I was hoping those with better knowledge of it would know for sure.  I don't think a None-type attribute exists in any PRPs, but I could be wrong.

It's really difficult to be certain about changes in versioned data files when MOULa shares versions with other MOULas.

core/PRP/Modifier/plPythonFileMod.h
Deledrius: Do we have an existing mechanism for aliasing them?

PR #270, opened 2023-03-28
Improve type hints.

Hoikas:
This adds two improvements to our PyHSPlasma type hints:
1. Instead of using the anti-pattern `Union[plKey, None]`, we now use `Optional[plKey]` when `None` is acceptable.
2. Added generic typing for `plKey` such that the `hsKeyedObject` subclass can be specified by, eg, `sceneObject: plKey[plSceneObject]`, allowing IDEs to know that `sceneObject.object` is a `plSceneObject`. Note that recursive type hinting (ie `sceneObject.key.object`) works on Python 3.11 and later.

Since a picture is worth 1,000 words, this is what you're getting:
image


zrax (approved):
Nice improvements, thanks!

Python/PyHSPlasma.pyi
```diff
@@ -661,7 +663,10 @@ class hsKeyFrame:
     def write(self, stream: hsStream) -> None: ...
 
 class hsKeyedObject(plCreatable):
-    key: plKey = ...
+    if sys.version_info.major == 3 and sys.verion_info.minor < 11:
+        key: plKey = ...
+    else:
+        key: plKey[Self] = ...
```
dgelessus, line 669: I know I'm late, but the Python version check can be written shorter (and more correctly) as:

```suggestion
    if sys.version_info < (3, 11):
        key: plKey = ...
    else:
        key: plKey[Self] = ...
```

And I *think* you can even make this fully compatible with older Python versions if you write it using `@property` and a regular type variable:

```suggestion
    @property
    def key(self: T) -> plKey[T]: ...
```

Hoikas: This LGTM. I think you should open this as a pull request 👍

PR #274, opened 2023-06-20
Clean up some unnecessary virtual methods

zrax:
Among other things, this fixes a cppcheck warning about calling virtual functions in constructors and destructors.


Python/Stream/pyStream.cpp
```diff
@@ -388,7 +382,6 @@ PY_METHOD_VA(Stream, writeLine, nullptr)
 }
 
 static PyMethodDef pyStream_Methods[] = {
-    pyStream_close_method,
```
Hoikas, line 391: This probably requires a corresponding update to the type hints for `hsFileStream` and `hsStream` in PyHSPlasma.pyi.

zrax: Ah, good point

Hoikas (approved):

PR #280, opened 2023-10-19
Really preserve ObjIDs.

Hoikas:
Previously, when a uoid is read in from the stream, the ObjID from the stream is discarded and reassigned to the value we already know. If the key is not already known, then the next sequential ID is given. If we want to preserve IDs, then we probably want to preserve *all* IDs, even those mid-stream. This is most apparent when diffing PRPs. IDs of zero are special values to the game client that mean "always look this up by name." This situation exists in the wild, right now, on MOULa's BahroCave_District_YeeshaCave.prp. When this file is read, libHSPlasma assigns sequential IDs to the `plMipmap` references in the layers if the Textures PRP is not already loaded. If the common PRP_as_Text.py is used to diff the PRP, a comparison is made against a random/junk value, resulting in a spurious looking change. See https://github.com/H-uru/moul-assets/pull/244 for an example of the confusion.


core/ResManager/plResManager.h
```diff
@@ -106,6 +106,24 @@ class HSPLASMA_EXPORT plResManager
...
 
+    /**
+     * Set whether or not keyed object IDs should be preserved from
+     * PRPs that are read in.  By default, key IDs are reordered to be
+     * consecutive when a PRP is read.  This sitting doesn't have any
+     * effect on data already loaded.
+     * \param preserve Preserve keyed object IDs
+     * \sa getPreserveObjIDs()
+     */
+    void setPreserveObjIDs(bool preserve) { preserveIDs = preserve; }
```
dpogue, line 117: FWIW, the original reason I didn't expose this except in the constructor was to ensure we couldn't end up in a situation where we had files open with both preserved key IDs and resequenced key IDs in the same ResMgr session.

dpogue (approved):

core/ResManager/plResManager.h
Hoikas: Yeah, I was thinking that as well - but I noticed that `setVer` already had a similar disclaimer, so I felt this was OK.... especially considering it would require re-allocating the res manager it `pyResManager__init__()` if this function didn't exist.

core/ResManager/plResManager.h
```diff
@@ -106,6 +106,24 @@ class HSPLASMA_EXPORT plResManager
      */
     void setVer(PlasmaVer pv, bool force = false);
 
+    /**
+     * Set whether or not keyed object IDs should be preserved from
+     * PRPs that are read in.  By default, key IDs are reordered to be
+     * consecutive when a PRP is read.  This sitting doesn't have any
+     * effect on data already loaded.
```
Deledrius, line 113:
```suggestion
     * consecutive when a PRP is read.  This setting doesn't have any
     * effect on data already loaded.
```

Deledrius (changes requested):

Deledrius (approved):

PR #281, opened 2024-03-02
Fix base class of `plLayerSDLAnimation`

dgelessus:
According to the MOULa source code, the correct superclass is `plLayerAnimationBase` and not `plLayerAnimation`. libHSPlasma largely seems to agree - the `CREATABLE` declaration and all of the super method calls already use `plLayerAnimationBase`. The base class declarations saying `plLayerAnimation` were probably by accident.

Someone who knows more than me about old Plasma should double-check this, but I think this change is correct for all Plasma versions - the read/write code has always used `plLayerAnimationBase` as the base class and that apparently worked fine for more than a decade.


zrax (approved):
This looks correct for PotS as well

PR #284, opened 2024-03-28
Fix plEAXListenerMod.h to be includable from outside libHSPlasma itself

dgelessus:
The header PRP/Audio/plEAXListenerMod.h included an EFX header from 3rdPartyLibs, which isn't part of the public libHSPlasma headers that get installed into the prefix. This made it impossible for downstream projects (e. g. PlasmaShop) to use plEAXListenerMod.h, because the required EFX header is only accessible within libHSPlasma itself.

This PR solves the issue by adding a new header plEAXStructures.h, which contains a copy of the few EAX structures needed by plEAXListenerMod.h. (This is intentionally similar to [plEAXStructures.h in H-uru/Plasma](https://github.com/H-uru/Plasma/blob/HEAD/Sources/Plasma/PubUtilLib/plAudio/plEAXStructures.h).) This header doesn't depend on the OpenAL/EFX headers and can thus be included by downstream projects, but the definitions are compatible with the real OpenAL/EFX headers, so it *should* be safe to include both at once.

Though I wonder if libHSPlasma needs to use the real OpenAL headers at all. libHSPlasma doesn't actually use the OpenAL API - it only needs a couple of EAX structure definitions and some constants. Those could be easily replicated in a custom header, perhaps with the names changed to avoid any possible conflicts with the real OpenAL headers.

### DIFF
--- a/Python/PRP/Audio/pyEAXListenerMod.cpp
+++ b/Python/PRP/Audio/pyEAXListenerMod.cpp
@@ -19,6 +19,7 @@
 #include <PRP/Audio/plEAXListenerMod.h>
 #include "PRP/Modifier/pyModifier.h"
 #include "PRP/KeyedObject/pyKey.h"
+#include <3rdPartyLibs/AL/EFX-Util.h>
 #include <3rdPartyLibs/AL/efx-unofficial.h>
 
 PY_PLASMA_NEW(EAXListenerMod, plEAXListenerMod)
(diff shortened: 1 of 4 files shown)


Deledrius:
This would help for a branch I have in progress for adding the EAXListener modifiers to PlasmaShop, for which I had to add the full EFXUtil.h in order to have access to the necessary structures.

Frustratingly, the lack of a complete header in Plasma is the major remaining blocker in completing [Plasma#1000](https://github.com/H-uru/Plasma/pull/1000), as the required structures only exist in one of OpenAL's internal headers and it isn't a public interface (even though it originally was as part of EAX), and I don't want to copy the file wholesale into our repo.  Deciding how to handle this has prevented me from finishing the otherwise-complete work.

zrax (approved):

PR #285, opened 2024-03-28
Fix `plArmatureLODMod` writing PRC `Meshes` tag twice

dgelessus:
The base class `plArmatureModBase` already writes the `Meshes` tag, so `plArmatureLODMod` shouldn't write it a second time.


dpogue (approved):
LGTM

dgelessus:
> These appear to be duplicated because the stream I/O also has duplicate readers and writers for them...

Actually, nothing is duplicated in the stream data either - but
the code is really wacky here. There are two places that read the meshes (`plArmatureModBase::read` and `plArmatureLODMod::read`), but only the latter is actually executed. Note that (for Uru PRPs) `plArmatureMod::read` doesn't call the base class method `plArmatureModBase::read`, but instead skips one class up the chain to `plAGMasterMod::read`. So one of the two places where the meshes are read is always skipped. The same thing happens when writing.

The trouble was that the PRC read/write code replicated one half of this wacky logic (two classes reading/writing the same field), but not the other half (skipping one of those two implementations).

PR #286, opened 2024-03-29
Add missing public getters and setters to various classes

dgelessus:
These classes were already fully implemented, but had no public API beyond the basic read/write methods, so there was no way to interact with them outside libHSPlasma. This PR adds getters and setters where appropriate and converts some classes to public structs.

This is a somewhat random selection of classes (I was experimenting with PrpShop and these are the ones I stumbled across). There are probably still more places that could use the same treatment.


core/PRP/Geometry/plMorphDelta.h
```diff
@@ -20,30 +20,25 @@
 #include "PRP/plCreatable.h"
 #include "Math/hsGeometry3.h"
 
-class HSPLASMA_EXPORT plVertDelta
+struct HSPLASMA_EXPORT plVertDelta
```
zrax, line 23: Why were these changed to structs?

dgelessus: It was the simplest way to make all the fields `public`. This seemed to be the convention in libHSPlasma for things that only hold simple data values.

But if you like, I can change it back to `class` and/or wrap the fields in getters/setters. I don't have opinions on the details here, as long as there's *some* way to access the data in the class.

dgelessus:
I'm noticing now that the first few commits of this PR overlap with #253. The other PR should probably be merged first, then I can rebase this PR to drop the redundant changes.

core/PRP/Avatar/plArmatureMod.h
```diff
@@ -79,6 +88,33 @@ class HSPLASMA_EXPORT plArmatureMod : public plArmatureModBase
 protected:
     void IPrcWrite(pfPrcHelper* prc) HS_OVERRIDE;
     void IPrcParse(const pfPrcTag* tag, plResManager* mgr) HS_OVERRIDE;
+
+public:
+    plKey getDefaultMesh() const { return fDefaultMesh; }
+    ST::string getRootName() const { return fRootName; }
```
zrax, line 94: I would probably return strings as references, since they will potentially invoke a deep copy otherwise

dgelessus: Makes sense. Most existing getters return `ST::string` by copy though (which is why I did the same thing initially).

PR #287, opened 2024-04-07
Fix handling of waveset bounding boxes.

Hoikas:
See CWE's `plGeometrySpans::AdjustBounds()` for the source of this change. Wavesets are flattened at runtime to the water height and are perturbed along the world Z axis. The input mesh, OTOH, conforms to the bottom of the pool. This means that if we simply spit out the bounding box of the input mesh, the resulting bounding box will be too deep in the -Z direction and not contain any of the +Z displacement needed. This means that the waveset could vanish due to culling.

NOTE: A draft because I'd like to test this change against korman.


Hoikas:
After some basic testing, I think this is ready.

zrax (approved):

PR #290, opened 2024-09-21
Add building for Python 3.12, x64 in CI

dpogue:
Also update to current versions of GHA steps to avoid deprecation warnings.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -7,6 +7,8 @@ jobs:
     strategy:
       matrix:
         platform:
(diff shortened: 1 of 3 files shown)


.github/workflows/windows-ci.yml
```diff
@@ -7,6 +7,7 @@ jobs:
     strategy:
       matrix:
         platform:
+          - { image: windows-2022, generator: Visual Studio 17 2022, toolset: v143, cmake-arch: x64, arch: x64, python: "3.12", unity: ON, str: windows-x64-v143 }
```
Hoikas, line 10: Maybe we should also add an entry for 3.11?

dpogue: Updated to include 3.11

zrax (approved):

PR #291, opened 2024-12-06
Add partial bindings for `plJPEG` and `plPNG`.

Hoikas:
I only included the two most immediately useful things in Python - loading JPEGs and PNGs from a stream into a `plMipmap`.


zrax (commented):
From the macOS CI failure, it looks like the JPEG and PNG dependencies need to be made public on HSPlasma...  Either that, or they should be removed from the public headers.

zrax (approved):

PR #304, opened 2025-07-27
Remove windows-2019 runners.

Hoikas:
The windows-2019 runner was removed from GHA about a month ago. Also, we need to test against at least Python 3.13.


.github/workflows/windows-ci.yml
```diff
@@ -7,14 +7,14 @@ jobs:
     strategy:
       matrix:
         platform:
+          - { image: windows-2022, generator: Visual Studio 17 2022, toolset: v143, cmake-arch: x64, arch: x64, python: "3.13", unity: OFF, str: windows-x64-v143 }
```
dpogue, line 10: Do we want `unity: OFF` here? I guess same question for the ones further down...

Is the goal of having the option just to make sure that don't introduce inadvertent compile-time issues between unified and non-unified sources?

dpogue (approved):

zrax (approved):

PR #314, opened 2026-07-27
Fix EAX values being wrong on Linux x64

dpogue:
On x64 Windows with MSVC, `long` is 32-bits, but on other x64 systems, `long` is 64-bits. When these values in the EAXPARAMETERS structure are populated, a 32-bit unsigned int is read from the stream and cast to the structure member type. Because these are defined as `long` this means that some values that are intended to be 32-bit signed numbers are getting interpreted as 32-bit unsigned numbers because of expansion to the 64-bit member type.


Deledrius (approved):

core/PRP/Audio/plEAXStructures.h
```diff
@@ -15,19 +15,19 @@ typedef struct _EAXVECTOR
 #define EAXREVERBPROPERTIES_DEFINED
 typedef struct _EAXREVERBPROPERTIES
 {
-    unsigned long ulEnvironment;
+    unsigned int ulEnvironment;
```
zrax: If these need to be a specific size, we should probably use `uint32_t` etc. instead.

dpogue: That's what I was originally thinking, but it seems like the bulk of libHSPlasma doesn't use stdint types. I'm definitely happy to switch it over if that's what we prefer

Hoikas (changes requested):
The getters and setters also need to be updated in PyHSPlasma to match.

Deledrius (approved):

Hoikas (approved):

zrax (approved):

